Issue #51, opened 2018-12-12
Schema test errors


jpmckinney:
I started working to restore the test that was commented out in Travis and fixed some errors already:

* Bad casing of `Code`, etc. in contractNature.csv
* Invalid JSON in the two test files
* Bad line endings in CSV files (`LFCR` or `CRLF` instead of `LF`)
* flake8 and whitespace

However, some issues require discussion:

* [x] 1. It's not specified whether contractingProcessStatus.csv or method.csv are open or closed codelists. I assume closed.
* [x] 2. We allow null in many fields, but this schema doesn't support merging. Therefore, I think we should remove null as an allowed value for all fields. I can do this via borrow-schema.py
* [x] 3. Should ID fields be required on objects in arrays? If so we're missing `Classification`, `Location`, `ContractingProcess`, `LinkedRelease`

Some missing properties:

* [x] /properties/budget/properties/amount/title
* [x] /properties/budget/properties/amount/description
* [x] /definitions/ContractingProcess/properties/summary/title
* [x] /definitions/ContractingProcess/properties/summary/description

I'll go with my assumptions in (1) in a PR, since I already started on those changes, but we can of course change the PR.

jpmckinney:
PR is #52

duncandewhurst:
Discussed with @jpmckinney:

2. is correct and will be included in alpha-2 - @jpmckinney will update the PR
3. will require more detailed review so we will wait until the beta release.

I'll address the missing properties

duncandewhurst:
#52 is merged, moving this issue to the beta milestone

duncandewhurst:
>     * [ ]  3. Should ID fields be required on objects in arrays? If so we're missing `Classification`, `Location`, `ContractingProcess`, `LinkedRelease`

I've reviewed whether these objects need required identifiers and propose the following:

* `ContractingProcess` -  we don't currently have an identifier on the contracting process object. We could make `contractingProcess/Summary/ocid` a required field, but that would which would require OC4IDS publishers to mint `ocid`'s when no OCDS data is available, and dealing with a nested identifier would have implications for tooling, in particular for conversion from spreadsheet to JSON format. Instead, I propose we add a required 'id' field on the parent object, which should be populated with the `ocid` when available otherwise a local identifier for the contracting process can be minted. We could then either remove the `ocid` field from the summary object, or leave it there but only have it used when an actual ocid exists (I prefer removing it).

[...]

PR #52, opened 2018-12-12
Fixes to project schema

jpmckinney:
- contractingProcessStatus.csv and method.csv are closed codelists. #51
- Arrays of strings should not allow null.
- Add enum for release tag.
- Have consistent order of codelist properties.


jpmckinney:
I updated the script to remove null, as well as `patternProperties` (to my knowledge, OC and Infrastructure doesn't use these).

jpmckinney:
Noting that we should merge master and re-run `python schema/borrow-schema.py` if any of the PRs touching the schema are merged first.

PR #53, opened 2018-12-12
environmentalImpact is already in documentType.csv

jpmckinney:


jpmckinney:
That's fine with me. Can you open a PR on the standard to do that? We can leave this PR open. Hopefully we'll be able to close it for the beta, but that depends on the review schedule for 1.1.4 (assuming the change in considered permissible in a patch release).

duncandewhurst:
We updated the description for the environmentalImpact code in https://github.com/open-contracting/infrastructure/pull/75

I've opened an issue in the OCDS repo to update the description there: https://github.com/open-contracting/standard/issues/818

I think this PR can be closed now, without merging.

PR #45, opened 2018-12-11
Move variations to contractingProcessSummary, rename fields, add old/new values and periods

duncandewhurst:

PR #57, opened 2018-12-19
update IDS mapping, add OCDS mapping

### LINKED ISSUE #34: Input mapping review
I was having trouble commenting in Google Sheets, so I am copying the task here. I'm assuming only the first sheet is relevant; let me know if I should be looking at any others.

> please review the OCDS input mapping I have added to IDS project level mapping table.

> This is based on the mapping you previously documented to OCDS 1.0RC, with updates to reflect changes since then and with some additional fields mapped (e.g. projected and actual cost at completion).

> I've used the language of 'infer' to indicate that, e.g. the buyer of the contract might not necessarily be the project owner, and plan to add some further prose to that effect on the page where the mapping appears.

> We could be more explicit about whether each element should be inferred from just the main construction contract (such as for the project location) or from all related contracts (such as for the total projected value of the project), but I've tried to keep the mapping concise and propose instead providing some guidance in prose on the page where the mapping appears.

duncandewhurst:
Closes #34 

Includes updates to reflect changes in #45 and #54, need to merge #45 first

Build is current failing, see https://github.com/open-contracting/standard-maintenance-scripts/issues/138


docs/cost/index.md
```diff
@@ -33,30 +33,73 @@ You can read more about the Infrastructure Data Standard in [CoST Guidance Note
...
+
+* How existing OCDS data can be used to populate project-level and contracting process summary data, in the 'Mapping from OCDS' column.
+
+```eval_rst
+.. admonition:: Mapping from OCDS
+    :class: Note
+
+    .. markdown::
+
+        Some mappings use fields from [OCDS extensions](http://standard.open-contracting.org/latest/en/extensions/), in these cases the name of the extension is noted in brackets and, where possible, alternative mappings are provided which use only fields from the core OCDS schema.
```
jpmckinney: Run-on sentence. brackets -> parentheses ('brackets' typically refers to square brackets). Inconsistent pluralization. Fixed version:

> Some mappings use fields from [OCDS extensions](http://standard.open-contracting.org/latest/en/extensions/). In these cases, the names of extensions are noted in parentheses; where possible, alternative mappings are provided that use only fields from the core OCDS schema.

jpmckinney:
I fixed the build.

PR #59, opened 2018-12-19
Adding example data

pindec:
This has been checked to have no additional fields against the schema in [5-variations](http://standard.open-contracting.org/infrastructure/5-variations/en/_static/project-level/project-schema.json) using the [Compile To JSON Schema tool](https://compiletojsonschema.readthedocs.io/en/latest/cli.html).


jpmckinney:
I fixed the build.

jpmckinney:
I haven't checked the example, but this looks fine. Feel free to merge this, and it will automatically close #45.

jpmckinney:
Actually, see my two comments in #45. If you can make those commits here, then we can merge – pending @duncandewhurst's review.

jpmckinney:
Was code used to generate the example, or was it all hand-edited? If there's code, we might want to store it, in order to re-generate it more easily if the schema changes.

duncandewhurst:
A blank template was generated using https://github.com/open-contracting/json-schema-random but other than that editing was done by hand

PR #63, opened 2018-12-20
Alpha 2 fixes

duncandewhurst:


docs/guidance/publishing-from-portal.md
```diff
@@ -50,7 +50,7 @@ Your mapping might identify:
 
 * **Gaps in your data** where data in the project-level data specification is not currently collected or disclosed in your system. Use the specification as a guide to the information that is important to users and consider whether your system and business processes could be updated to collect and publish additional information.
 
-* **Gaps in the data specification** where data is collected by your system but doesn’t map to the specification. Rather than being excluded from your publication, such information should be included as additional fields in your data. Refer to [extending the schema](../projects/schema.md#extending-the-schema) for information on including additional fields in your data.
+* **Gaps in the data specification** where data is collected by your system but doesn’t map to the specification. Rather than being excluded from your publication, such information should be included as additional fields in your data. Refer to [extending the schema](../projects/schema.md) for information on including additional fields in your data.
```
jpmckinney: @duncandewhurst Why were the fragment identifiers removed here and for `reference.md#contractingprocess` above? Without them, users need to scroll more.

duncandewhurst: they were causing broken links, but I figured out the correct syntax now

duncandewhurst: See https://github.com/open-contracting/infrastructure/pull/66

PR #75, opened 2019-01-30
merge ocds, ppp and infra codelists

### LINKED ISSUE #25: Review and merge document type codelists
Currently the `schema/project-level/codelists` directory contains two documentType codelists:

* `+documentType.csv` which contains the new codes introduced in the project level data specification
* `documentType.csv` which contains all the codes from the OCDS for PPPs profile

Not all codes from OCDS for PPPs are relevant to the project level data specification and we should aim to have a single documentType codelist in the beta, so the two lists will need reviewing and merging.

duncandewhurst:
closes #25 

see comment in issue for methodology

### DIFF
--- a/schema/project-level/codelists/+documentType.csv
+++ b/schema/project-level/codelists/+documentType.csv
@@ -1,7 +1,0 @@
-Code,Title,Description
-projectScope,Project scope,"A description of the main outputs from the project that are being taken forward into construction (including type, quantity and units)"
-landAndSettlementImpact,Land and Settlement Impact,"State the amount of land and property that was acquired for the project (e.g. 25sq km land), and outline related impacts (e.g. archaeological issues (moved saxon burial site), local/indigenous settlements (relocated 5 indigenous villages of 500 villagers each), impacts on local businesses e.g. (30 business properties purchased))."
-environmentalImpact,Environmental Impact,"Describe the environmental impacts and
-mitigation measures for this project (e.g. impacts on flora, fauna & woodlands, areas of natural beauty, carbon emissions etc.) and mitigation measures (e.g. pollution control, low carbon solutions, sustainable timber etc.)"
-projectEvaluation,Project evaluation,"Generally published at the conclusion of a project, providing a technical and financial summary of delivery."
-budgetApproval,Budget approval,"Additional details about the approval of the budget."
--- a/schema/project-level/codelists/documentType.csv
+++ b/schema/project-level/codelists/documentType.csv
@@ -1,5 +1,5 @@
 Code,Title,Description,Source
-needsAssessment,Needs Assessment,"Documentation of the needs assessments carried out for this contracting process or project addressing demand for the project or investment from the affected communities or users.",OC for Infrastructure
+needsAssessment,Needs Assessment,"Documentation of the needs assessments carried out for this contracting process or project addressing demand for the project or investment from the affected communities or users.",OC4IDS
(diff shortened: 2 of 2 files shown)


jpmckinney:
Hmm, the changes to titles/descriptions raise some issues. We allow the PPPs profile to make some incompatible changes (like removing fields), but I'm not sure if we want to carry all of those over here.

Specifically:

Code | Title | Description | OCDS Title (if different) | OCDS Description (if different)
-- | -- | -- | -- | --
needsAssessment | Needs Assessment | Information on the project need, benefits provided, and economic and social impact |   | Documentation of the needs assessments carried out for this contracting process addressing demand for the project or investment from the affected communities or users.
contractSchedule | Schedules and milestones | Documentation setting out key milestones and schedules. Structured versions of this information may be modelled using the milestones block. | Contract schedules | Any document which contains additional terms, obligations or information related to the contract, such as a schedule, appendix, annex, attachment or addendum.

The semantics of `needsAssessment` are changed from being about the contracting process to being about the project. The semantics of `contractSchedule` are changed from being about "contract schedules" in a legal sense (annexes, etc.) to being about schedules in a time sense (milestones, etc.).

The PPPs profile should have created new codes for these, rather than changing semantics. (I'll create an issue there.) For the Infrastructure work, we should keep the OCDS titles/descriptions; we can create the new codes that the PPPs profile should have created if we want to keep the alternate semantics. Does that sound reasonable?

jpmckinney:
This project changes the description of `environmentalImpact` from "Documentation of environmental impact assessments carried out before, during or after the contracting process." to "Describe the environmental impacts and mitigation measures for this project (e.g. impacts on flora, fauna & woodlands, areas of natural beauty, carbon emissions etc.) and mitigation measures (e.g. pollution control, low carbon solutions, sustainable timber etc.)".

It changes the semantics from being about the contracting process to being about the project.

I think we should either change the description to be agnostic to what it is describing ("the contracting process or project" or "the object this document is about"), or have a separate codelist for project documents.

jpmckinney:
Noting that we can/ought to make some changes to `borrow-schema.py` to make this codelist building reproducible (e.g. if we add more codes to OCDS or PPPs).

duncandewhurst:
Thanks @jpmckinney 

For `needsAssesment` I've updated the description to be contract/project agnostic.

For `contractSchedules` I've updated the title and description to match OCDS core.

For `environmentalImpact` I've updated the description to be contract/project agnostic.

jpmckinney:
The updated `needsAssessment` adds "or contracting process", but otherwise uses the re-definition from PPPs, which somewhat confusingly adds "economic and social impact" into it. I prefer to add "project or" to the original OCDS definition.

I'd prefer closer alignment between the PPP and Infra definitions of `environmentalImpact`. The Infra definition is an imperative sentence, which is not how the first sentence of a code description is structured elsewhere. Given that we can change PPPs easily, we can update both given that neither description is best.

duncandewhurst:
I've updated the description for `needsAssessment` per your suggestion.

For `environmentalImpact`, how about the following definition (based on both the PPP and infrastructure codes):

> Documentation of assessments of the environmental impacts (e.g. impacts on flora, fauna & woodlands, areas of natural beauty, carbon emissions etc.) and mitigation measures (e.g. pollution control, low carbon solutions, sustainable timber etc.) for this contracting process or project.

We could also add "carried out before, during or after the contracting process or project." from the OCDS for PPPs description, but I don't think it's necessary.

jpmckinney:
Looks good!

I agree – the extra clause from PPPs doesn't clarify the concept.

PR #93, opened 2019-02-26
43 party role codelist

### LINKED ISSUE #43: Review party role codelists
Relates to items 1. and 5. in https://github.com/open-contracting/infrastructure/issues/35

Currently OC for Infrastructure re-uses the the party role codelist from OCDS for PPPs, however the output mapping suggests using several codes from the core OCDS party role codelist ('funder' from #35 plus 'supplier' in the process level mapping)

Task is to review the mapping and the two codelists, determine which codes are required for OC for Infrastructure and which codelist is most appropriate.

duncandewhurst:
Closes #43


duncandewhurst:
See comment in #43 for rationale for changes

PR #97, opened 2019-03-05
add project identifier guidance

### LINKED ISSUE #58: Global uniqueness of project identifiers
The project level schema asks publishers to provide:

> A unique identifier or reference number for this infrastructure project.

Currently there's no further guidance or mechanism to prevent multiple publishers using the same identifier for an infrastructure project.

In OCDS an [ocid prefix](http://standard.open-contracting.org/latest/en/schema/identifiers/#contracting-process-identifier-ocid) is assigned to publishers by the OCDS helpdesk and prepended to existing identifiers for contracting processes to make them globally unique.

Options for ensuring global uniqueness of project identifiers in OC for Infrastructure:

1. Do nothing - risk of identifier clash for infrastructure projects from different publishers

2. Specify an identifier format based on the publisher name/organization identifier - I don't think we should pursue this option for the same [reasons](https://github.com/open-contracting/standard/issues/73#issuecomment-57015793) it was discounted in OCDS

3. Require publishers to request an OCID prefix from the OCDS helpdesk and prepend this to their project identifier - risk of identifier clash if infrastructure projects and contracting processes share an identifier

4. Require publishers to request project identifier prefix from the OCDS helpdesk (i.e. helpdesk maintains a separate list of prefixes for OC for Infrastructure publishers)

5. Require publishers to generate a [UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier) to prepend to their infrastructure project identifiers - this approach is used in the [draft Beneficial Ownership Data Standard](http://standard.openownership.org/en/schema-beta-2/schema/guidance/statement-identifiers.html#strategies-for-identifier-creation).

@jpmckinney @lindseyAM - what do you think?

duncandewhurst:
Closes #58 

This PR adds guidance on project identifiers based on the discussion in #58 and the [blog](https://www.open-contracting.org/2018/10/03/exploring-project-identifiers-for-infrastructure-projects/) published following the research phase of the project


docs/guidance/identifiers.md
```diff
@@ -0,0 +1,38 @@
+# Project identifiers
+
+A project identifier is a globally unique identifier for an infrastructure project. Every project in OC4IDS has a project identifier in the `id` field.
+
+Project identifiers can be used to join up information published at different times or from different systems, for example, including a project identifier in contracting data enables information on the design, construction and supervision contracts for a single infrastructure project to be joined up.
```
jpmckinney: Use semi-colon instead of comma before "for example".

jpmckinney: > including a project identifier in contracting data makes it possible to join up data on the design, construction and supervision contracts within a single infrastructure project.

Move verb before object.

jpmckinney: Use 'data' not 'information' in first sentence.

docs/guidance/identifiers.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Project identifiers in contracting data
+
+With an authoritative list of infrastructure projects in a country to draw on, and if the identifiers it assigned are used within contracting systems, it would be possible to automatically check which projects have published contracting information, and to analyse that data to identify projects where further scrutiny, monitoring and data collection should take place.
```
jpmckinney: > A common need is to access data about the contracting processes related to an infrastructure project. When contracting systems use consistent identifiers to refer to infrastructure projects, this becomes possible. An example use case is automatically checking which projects have related contracting data, and then manually filtering projects for further scrutiny, monitoring or data collection.

Avoid mixing "with…" and "if…" clauses and unmatched verb tenses. Use 'data' instead of 'information'. Use American English for 'analyze'. State the two examples as examples.

This sentence was trying to do a lot. Using three sentences makes it easier to parse. The key concept is consistent identifiers; authoritative lists are one means to achieve that.

docs/guidance/identifiers.md
```diff
@@ -0,0 +1,38 @@
...
+
+There are different approaches to including project identifiers in contracting data, with the best solution depending on the context of an implementation:
+
+* **Include an open field for project identifiers in procurement systems** and work with each department entering data to make sure this is populated according to some defined pattern.
```
jpmckinney: We typically use 'free-text' instead of 'open'.

'department entering' -> 'group that enters'

docs/guidance/identifiers.md
```diff
@@ -0,0 +1,38 @@
...
+
+  This would allow some checks to be run, for example, checking that all the contracting processes over a certain value from a given agency have a project identifier, and that the identifier matches a known pattern or a local identifier list.
```
jpmckinney: In this and the next paragraph, remove the modal "would" (i.e. simply "allows") in the first sentence.

Use colon before "for example". Remove "known" as this is implied.

docs/guidance/identifiers.md
```diff
@@ -0,0 +1,38 @@
...
+
+  This would also allow data on several contracting processes relating to a single project to be joined up, however allowing free-text entry of project identifiers would likely lead to issues with data quality and could result in identifier clash between projects from different agencies or departments.
```
jpmckinney: > Allowing free-text entry of project identifiers can lead to data quality issues; for example, two groups can accidentally choose to use the same identifier for different projects.

I think we should lead with the caveats, and then say what the pattern makes possible.

docs/guidance/identifiers.md
duncandewhurst: fixed in https://github.com/open-contracting/infrastructure/pull/97/commits/6dbddcc20150fb09e92a8ab533ffde539d1f8ae7

docs/guidance/identifiers.md
duncandewhurst: fixed in https://github.com/open-contracting/infrastructure/pull/97/commits/6dbddcc20150fb09e92a8ab533ffde539d1f8ae7

docs/guidance/identifiers.md
duncandewhurst: fixed in https://github.com/open-contracting/infrastructure/pull/97/commits/6dbddcc20150fb09e92a8ab533ffde539d1f8ae7

docs/guidance/identifiers.md
duncandewhurst: fixed in https://github.com/open-contracting/infrastructure/pull/97/commits/6dbddcc20150fb09e92a8ab533ffde539d1f8ae7

docs/guidance/identifiers.md
duncandewhurst: fixed in https://github.com/open-contracting/infrastructure/pull/97/commits/6dbddcc20150fb09e92a8ab533ffde539d1f8ae7

docs/guidance/identifiers.md
```diff
@@ -0,0 +1,40 @@
...
+
+Project identifiers can be used to join up data published at different times or from different systems; for example, including a project identifier in contracting data makes it possible to join up data on the design, construction and supervision contracts within a single infrastructure project.
+
+## Project identifiers in contracting data
+
+A common need is to access data about the contracting processes related to an infrastructure project. When contracting systems use consistent identifiers to refer to infrastructure projects, this becomes possible. An example use case is automatically checking which projects have related contracting data, and then manually filtering projects for further scrutiny, monitoring or data collection.
+
+There are different approaches to including project identifiers in contracting data, with the best solution depending on the context of an implementation:
+
+* **Include a free-text field for project identifiers in procurement systems** and work with each group that enters data to make sure this is populated according to some defined pattern.
```
jpmckinney: Change 'each group that enters data' to 'officials entering procurement information' to be consistent with next solution.

docs/guidance/identifiers.md
```diff
@@ -0,0 +1,40 @@
...
+
+  Free-text entry of project identifiers can lead to data quality issues; for example, two groups can accidentally choose to use the same identifier for different projects.
+
+  However, this approach allows some checks to be run; for example, checking that all the contracting processes over a certain value from a given agency have a project identifier, and that the identifier matches a pattern or a local identifier list.
+
+  This approach also allows data on several contracting processes relating to a single project to be joined up.
+
+
```
jpmckinney: Remove extra newline.

docs/guidance/identifiers.md
```diff
@@ -0,0 +1,40 @@
...
+* **Establish a national project register managed by a central agency and integrated into procurement systems.** In this model, officials entering procurement information would look up a project from the national register to include its identifier in contracting data, and would have to request its addition if it was not already in the register.
```
jpmckinney: > would look up and use the project's identifier from the national register. If the project is not yet in the register, they would request its addition.

docs/guidance/identifiers.md
```diff
@@ -0,0 +1,40 @@
...
+
+  This approach allows validation ‘at source’ of the project identifier and would support checks on how many projects have no associated contracts, however it would require the availability of a central actor, the political will and the technical capacity to establish and maintain a register.
```
jpmckinney: allows -> enables

would support -> supports, though the check mentioned here can be done under the other approach, too…

Semi-colon before "however", though the "however" clause is a bit high-level / obvious, so I'd be in favor of removing it.

I'm not sure that the current text demonstrates the advantage of a central registry. Can a sentence be added about avoid the data quality issues from the other approach?

docs/guidance/identifiers.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Globally unique identifiers
+
+The project identifier should be globally unique, this means that across data from all OC4IDS publishers, it should refer to one specific infrastructure project.
```
jpmckinney: > ; this means that, across the data of all OC4IDS publishers, each project identifier refers to exactly one infrastructure project.

docs/guidance/identifiers.md
```diff
@@ -0,0 +1,40 @@
...
+
+To make the project identifier globally unique the OC4IDS helpdesk assigns a prefix which publishers must add to existing project identifiers, using the following identifier structure: `[project identifier prefix]-[existing project identifier]`
```
jpmckinney: Add comma after unique. Capitalize Helpdesk (also in admonition).

which -> that

It's not clear what an "existing project identifier" is (especially in contexts where there isn't a practice of assigning identifiers). A sentence (or more) should be added to relate this section to the previous section. We might want to use the phrase "local project identifier" in this section and the previous section, and to describe, in this section, that a prefix is added to a local project identifier to make it globally unique.

We might also want to describe, in the previous section, that a project identifier should be locally unique, in a similar way to the first paragraph in this section.

docs/guidance/identifiers.md
duncandewhurst: Updated in https://github.com/open-contracting/infrastructure/pull/97/commits/acaa4b9fc84a4b1e1fba14ac3e30d85bda1a46f0

docs/guidance/identifiers.md
duncandewhurst: Fixed in https://github.com/open-contracting/infrastructure/pull/97/commits/acaa4b9fc84a4b1e1fba14ac3e30d85bda1a46f0

docs/guidance/identifiers.md
duncandewhurst: Updated in https://github.com/open-contracting/infrastructure/pull/97/commits/acaa4b9fc84a4b1e1fba14ac3e30d85bda1a46f0

docs/guidance/identifiers.md
duncandewhurst: Updated in https://github.com/open-contracting/infrastructure/pull/97/commits/acaa4b9fc84a4b1e1fba14ac3e30d85bda1a46f0

I added 'more comprehensive' to the check to signal the improvement over the free-text approach.

I prefer to keep the however clause, as I think it's a useful pointer to wider considerations on implementing project identifiers (i.e. it's not just a technical/data issue). In particular I think it is good reminder to helpdesk analysts and other support providers to flag these considerations to publishers.

I added a sentence highlighting how the data quality issues can be avoided.

docs/guidance/identifiers.md
duncandewhurst: Updated in https://github.com/open-contracting/infrastructure/pull/97/commits/acaa4b9fc84a4b1e1fba14ac3e30d85bda1a46f0

docs/guidance/identifiers.md
duncandewhurst: I've tried to address your suggestions in https://github.com/open-contracting/infrastructure/pull/97/commits/acaa4b9fc84a4b1e1fba14ac3e30d85bda1a46f0

Happy for you to refine further by committing directly to the branch

jpmckinney:
I made two commits, feel free to merge.

Issue #91, opened 2019-02-26
Update font for headings and menu items to Monserrat 


duncandewhurst:
https://fonts.google.com/specimen/Montserrat

duncandewhurst:
Closed by #100

PR #100, opened 2019-03-05
[#91] New font added to menus and headings.

kindly:


duncandewhurst:
thanks @kindly, can we also apply the font to the heading at the top of the menu?

image

kindly:
@duncandewhurst that is done now.

PR #105, opened 2019-03-08
51 object identifiers

### LINKED ISSUE #51: Schema test errors

duncandewhurst:
Closes #51

### DIFF
--- a/schema/project-level/project-schema.json
+++ b/schema/project-level/project-schema.json
@@ -563,6 +563,12 @@
       "title": "Delivery Location",
       "description": "The location where activity related to this project will be delivered, or will take place. A location may be described using a geometry (point location, line or polygon), a gazetteer entry, an address, or a combination of these.",
       "properties": {
+        "identifier": {
+          "title": "Identifier",
+          "description": "A local identifier for this location, unique within the array this location appears in.",
+          "type": "string",
+          "minLength": 1
+        },
         "description": {
           "title": "Description",
           "description": "A name or description of this location. This might include the name(s) of the location(s), or might provide a human readable description of the location to be covered. This description may be used in a user-interface.",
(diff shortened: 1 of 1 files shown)


schema/project-level/project-schema.json
```diff
@@ -554,6 +569,12 @@
       "title": "Delivery Location",
       "description": "The location where activity related to this project will be delivered, or will take place. A location may be described using a geometry (point location, line or polygon), a gazetteer entry, an address, or a combination of these.",
       "properties": {
+        "identifier": {
```
jpmckinney: Shouldn't this be `id`?

duncandewhurst: Fixed in https://github.com/open-contracting/infrastructure/pull/105/commits/53317e432480bc5fa8bee7e63cc1301fd0f7664d

PR #104, opened 2019-03-06
Documentation updates from the latam training call #95

pindec:
To address issues in issue #95

PR #106, opened 2019-03-08
refine language, remove newline, fix indentation

duncandewhurst:
Addresses review comments from #104


jpmckinney:
See https://github.com/open-contracting/infrastructure/pull/104#pullrequestreview-211975311

docs/projects/index.md
```diff
@@ -4,24 +4,23 @@ The regular disclosure of structured data can greatly enhance the transparency a
...
 ```eval_rst
 .. admonition:: Tip
     :class: tip
 
     .. markdown::
 
-      The term 'project' can also be used to refer to a larger programme of work of which an individual infrastructure project is a part, such as the construction of a whole highway. The term 'project' may also refer to a public investment project with funds allocated in the national budget, such as the upgrading of the rail network in a particular region.
-
-      However, in OC4IDS, the term 'project' refers to the construction of a single asset, for example the construction of a bridge, a highway junction or building.
+      In the context of OC4IDS, the term 'project' is defined as the delivery of a set of infrastructure assets in a specified location, generally under a single managing entity (project owner) and budget authority: for example, the construction of a bridge, highway junction or building.
```
jpmckinney: To which terms do "single managing entity (project owner) and budget authority" relate within the project schema? If there is a correspondence, it would be clearer to reuse the same terms.

I don't like that guidance note 6 defines it as "the delivery of", as we know a project is not one phase… In the documentation, we can abbreviate to "the term 'project' refers to a set of …"

Given the new definition, we should update the examples, which are all single assets. We can update it to a single asset and a set of assets (perhaps "; for example, a highway overpass or a university campus").

schema/project-level/project-schema.json
```diff
@@ -2,7 +2,7 @@
   "$schema": "http://json-schema.org/draft-04/schema#",
   "id": "https://github.com/open-contracting/infrastructure/blob/master/schema/schema.json",
   "title": "Open Contracting for Infrastructure Data Standard Schema",
-  "description": "The Open Contracting fo Infrastructure Data Standard Schema sets out a data structure for capturing and exchanging information about infrastructure projects in line with the disclosure requirements of the [CoST Infrastructure Data Standard](http://infrastructuretransparency.org/resource/977/).",
+  "description": "The Open Contracting for Infrastructure Data Standard Schema sets out a data structure for capturing and exchanging information about infrastructure projects in line with the disclosure requirements of the [CoST Infrastructure Data Standard](http://infrastructuretransparency.org/resource/977/). In the context of OC4IDS, the term 'project' is defined as the delivery of a set of infrastructure assets in a specified location, generally under a single managing entity (project owner) and budget authority: for example, the construction of a bridge, highway junction or building.",
```
jpmckinney: Update according to earlier comment.

docs/projects/index.md
jpmckinney: Discussed "a single procuring entity and ".

duncandewhurst: I've made these updates in https://github.com/open-contracting/infrastructure/pull/106/commits/bcef0830fc95a01edd2c4e49c8e53a42e7a6e5ee, I also replaced 'under' with the 'the responsibility of" 

However, on reflection I'm not sure removing 'delivery of' actually makes the description clearer for the reader, since the new definition of 'a set of infrastructure assets in a specified location' could refer to an already existing set of assets for which no further work is planned (rather than a project to create new assets or update existing assets).

Since it would be over long to list all the project phases (identification, preparation, implementation and completion) we could instead use a catch all term which isn't used elsewhere in the schema or documentation (unlike delivery), e.g.

> In the context of OC4IDS, the term 'project' refers to the development of a set of infrastructure assets in a specified location, generally the responsibility of a single procuring entity and budget authority: for example, a highway overpass or a university campus.

jpmckinney: That works for me. Feel free to make the change in the two locations and merge.

Issue #114, opened 2019-03-12
Update blank JSON file to include project package and object identifiers


duncandewhurst:
PRs #105 and #89 made changes which should be reflected in the blank JSON file

duncandewhurst:
Closed by #119

PR #119, opened 2019-03-15
updating example and blank json files

pindec:
- updated blank.json to have project package
- updated example.json to add more fields, and same project package order as blank.json ( and schema)


pindec:
Blank json updates related to #114

jpmckinney:
Diff without whitespace changes: https://github.com/open-contracting/infrastructure/pull/119/files?w=1

PR #125, opened 2019-03-19
Add custom version switcher

robredpath:
Figured it out! 

No need to merge this any time soon, but it means everything's here for when we want it.

### DIFF
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -10,3 +10,8 @@
     <img src="{{ pathto('_static/images/logos/') }}cost.png" style="padding-left:40px; padding-right:40px;"/>
   </a>
 {% endblock %}
+
+
+{% block version_options %}
+<!--#include virtual="/includes/version-options-infrastructure.html" -->
+{% endblock %}


duncandewhurst:
Thanks @robredpath 

The version switcher is working (showing live and master branches) but we'll want to merge this PR before deploying any further versions which should appear in the switcher.

robredpath:
Yeah, there's a sensible default that works for now. But, when we want to customise it, we can merge this and then update the relevant file in opendataservices-deploy.

docs/_templates/layout.html
```diff
@@ -10,3 +10,8 @@
     <img src="{{ pathto('_static/images/logos/') }}cost.png" style="padding-left:40px; padding-right:40px;"/>
   </a>
 {% endblock %}
+
```
jpmckinney, line 13: Remove extra newline, please :)

jpmckinney:
I'm not sure why I closed this PR (perhaps accidentally?), but I made an identical commit just now.

PR #166, opened 2019-10-29
add guidance on project identifiers in OCDS

### LINKED ISSUE #96: Clarify where project identifiers should be provided in OCDS data
@juanpane flagged that the `planning/budget/projectID` field in OCDS should be used to provide identifiers for national budget projects, not individual infrastructure projects.

We should determine whether the `planning/project/id` field introduced in the [budgets and projects extension](https://github.com/open-contracting-extensions/ocds_budget_projects_extension) would be appropriate or whether a new extension needs to be created.

Once the correct approach is decided we should include guidance in the OC4IDS documentation.

cc @jpmckinney

duncandewhurst:
* closes #96


docs/guidance/identifiers.md
```diff
@@ -29,6 +29,12 @@ There are different approaches to including project identifiers in contracting d
 
   However, this approach requires political will and technical capacity to establish the central register and integrate it into procurement systems, and it requires an appropriate central actor to manage it.
 
+### Project identifiers in OCDS
```
jpmckinney, line 32: Re-order this section by starting with what publishers should do, and then explain what they should not do… The first sentence of the current first paragraph can then more simply read:

> The `planning/budget/projectID` field should **not** be used for this.

In the current first paragraph, I'm confused as to whether 'project from the national budget' still means 'individual infrastructure project' or something else. This should be clarified.

duncandewhurst: See https://github.com/open-contracting/infrastructure/pull/166/commits/a85aec630c8769ea1d42dcf09319da7b1f04cf44, feel free to suggest further refinements or clarifications

duncandewhurst:
@jpmckinney were you happy with the changes here?

PR #168, opened 2019-10-29
152 add project type expansion

### LINKED ISSUE #152: Consider to add 'expansion' code to projectType
There are some infrastructure projects, like the airport expansions, who seeks to expand the existing assets.

It would be helpful if we can add the 'expansion' code to the projectType codelist, with the following description.

"The primary focus of this project is the expansion of an existing asset.

This code will be used at Nuevo León's implementation.

duncandewhurst:
Closes #152


pindec (commented):
Looks good to me.

jpmckinney (approved):

PR #169, opened 2019-10-29
fixes #158 contractingProcesses/releases/tag should be an array

### LINKED ISSUE #158: contractingProcesses/releases/tag is string instead of array
Should `contractingProcesses/releases/tag` be an array? Definition says 

> One **or more** values from the releaseTag codelist...

screenshot-standard open-contracting org-2019 10 21-15_05_40

pindec:
fixes #158


duncandewhurst (approved):

duncandewhurst:
@jpmckinney this would be a backwards incompatible change. Since we do not have any live publishers yet and since this was clearly an error, I think that is OK as long as we contact each implementer to let them know.

What do you think?

jpmckinney:
Yes, that's fine in an 0.x release. Please update the changelog to add backticks and add "(bugfix)" to the end of the changelog entry. Alternately, we can have a "bugs" sub-section like in the OCDS changelog: https://standard.open-contracting.org/latest/en/schema/changelog/

pindec:
We'll also need to update blank and example files.

pindec:
- Updated changelog with backticks and (bugfix)
- Updated example file (blank file will be re-generated once schema fixes all done)

PR #174, opened 2019-10-29
fixes #141 update descriptions of contractingProcesses/summary/modifications and summary/description

### LINKED ISSUE #141: Contracting processes summary: Update descriptions of `modifications` and `description`
The current description for `modifications` is:

> Details of changes to the duration, price, scope or other significant features of the contracting process should be logged here.
> 
> A semi-structured description of each change can be provided in the relevant description field, and the justification of the change given in reason.
> 
> When a modification is detected, the contracting process summary information should also be updated.

I think this needs two updates:

1. Remove the following statement:

> When a modification is detected, the contracting process summary information should also be updated.

Modifications can relate to either the value, duration or scope of the contracting process, however the `summary/contractValue` and `summary/contractPeriod` properties are the **initial** values, so shouldn't be updated.

We should also update the description of `summary/description` to make it clear this should be the initial scope of the contract.

2. Remove the following statement:

> A semi-structured description of each change can be provided in the relevant description field, and the justification of the change given in reason.

The modifications model now has structured fields for the before and after values, so we can remove this statement about providing semi-structured information.

pindec:
fixes #141


schema/project-level/project-schema.json
```diff
@@ -258,7 +258,7 @@
         },
         "description": {
           "title": "Description",
-          "description": "The description should summarise the purpose of this contract and the scope of work to be undertaken.",
+          "description": "The description should summarise the purpose of this contract and the initial scope of work to be undertaken.",
```
duncandewhurst: Somehow I think this makes it clearer that we mean the initial scope for the whole contract, rather than the scope of the initial work that will happen under the contract:

```suggestion
          "description": "The description should summarise the purpose of this contract and the initial scope of the work to be carried out under the contract.",
```

duncandewhurst (approved):
Happy to merge once the comment is addressed

jpmckinney:
I made the change and expanded the changelog entry.

jpmckinney (approved):

PR #176, opened 2019-10-29
fixes #132 add a publicAuthority organization reference field on project

### LINKED ISSUE #132: Add a `publicAuthority` organization reference field on project
Currently the project owner / public authority is only disclosed in the parties section.

To determine the owner of a project this means users need to check two arrays (`parties` and `parties/roles`).

We should consider adding a `publicAuthority` organization reference on the `project` object so that users and applications can more easily filter by the project owner.

pindec:
fixes #132


duncandewhurst (changes requested):
Please add to the worked example JSON file and guidance page.

pindec:
Note to also update blank.json

jpmckinney:
blank.json is generated by https://github.com/open-contracting/json-schema-random right?

pindec:
> blank.json is generated by https://github.com/open-contracting/json-schema-random right?

Yes. 

Is this on the deploy list of things to do anyway, so doing it manually at each schema change is unnecessary?

jpmckinney:
It wasn't, but I've now added it, so we can just do this once before release :)

pindec:
Brilliant, thanks @jpmckinney!

pindec:
@duncandewhurst I've added a new entry in parties for a publicAuthority, and added it to the publicAuthority field. A different party still has the roles of procuringEntity, buyer and administrativeEntity - is that realistic?

I've also updated the worked example page, but should I also add an entry in the "overview" list? It's a bit arbitrary what appears in there, but it's roughly the larger objects (which would discount adding publicAuthority).

duncandewhurst:
> @duncandewhurst I've added a new entry in parties for a publicAuthority, and added it to the publicAuthority field. A different party still has the roles of procuringEntity, buyer and administrativeEntity - is that realistic?

I would think it's most likely that the same party will have all 4 roles, so let's just add the 'publicAuthority' role to the existing party.

> I've also updated the worked example page, but should I also add an entry in the "overview" list? It's a bit arbitrary what appears in there, but it's roughly the larger objects (which would discount adding publicAuthority).

I think it's OK to leave it out of the overview list, as it appears later in the worked example anyway.

pindec:
> I would think it's most likely that the same party will have all 4 roles, so let's just add the 'publicAuthority' role to the existing party.

Thanks, I've added `publicAuthority` role to the existing party and removed the additional one originally added just for that role.

duncandewhurst (approved):

jpmckinney:
@duncandewhurst I added a commit (and fixed merge conflicts). Feel free to merge if no issue.

PR #185, opened 2019-11-14
Set all `id` fields to only permit strings

### LINKED ISSUE #184: Don't allow integer type for identifiers
Similar to https://github.com/open-contracting/standard/issues/789, OC4IDS also has some `id` fields which permit only strings and some which permit either strings or integers.

@jpmckinney given the early stage of implementations, should we fix this in 0.9.2 and make all `id` fields only permit strings?

duncandewhurst:
Closes #184

### DIFF
--- a/schema/borrow-schema.py
+++ b/schema/borrow-schema.py
@@ -100,6 +100,11 @@ def compare(actual, infra_list, ocds_list, prefix, suffix):
         sys.exit('{prefix} is missing {items}: remove from infra_{suffix} in borrow-schema.py?'.format(
             items=', '.join(removed), prefix=prefix, suffix=suffix))
 
+#set all `id` fields to only permit strings
+def remove_integer_identifier_types(schema):
+    for definition in schema['definitions'].values():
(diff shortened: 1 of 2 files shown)


duncandewhurst:
@jpmckinney I've taken a pass at adding this to `borrow-schema.py` as it seemed a good opportunity to get a better handle on how the script works

jpmckinney:
I just noticed that your shell must have interpreted `id` as a command, and put its output in the commit message. I tried to rebase, but there are so many merge conflicts (basically every PR so far has had merge conflicts) that I can't get it to a sound state.

duncandewhurst:
Ah, I also didn't notice that happened.

None of the information in the commit message is sensitive, so shall we just leave it?

Regarding the merge conflicts, I think these are mostly in the changelog - is there a better way of doing that to prevent the merge conflicts?

jpmckinney:
Yeah, there's nothing sensitive, so okay to leave.

The merge conflicts on individual PRs are fine – it was just a problem when rebasing so many PRs.

PR #187, opened 2019-11-15
add requestDate on budget

### LINKED ISSUE #154: Add 'requestDate' on 'budget' object
There is a use case for this extension. 

When having a central institution responsible for the contracting process of infrastructure projects, there is a need to inform the other institutions when the budget was requested and when the budget was finally approved. So, with this, the central institution can justify why the project has not been initiated.

You can see a draft of this extension here: https://github.com/antonioherrera/oc4ids_requestDate_extension

duncandewhurst:
Closes #154

### DIFF
--- a/docs/examples/example.json
+++ b/docs/examples/example.json
@@ -96,6 +96,7 @@
           "amount": 40000000,
           "currency": "GBP"
         },
+        "requestDate": "2015-05-30T00:00:00Z",
         "approvalDate": "2015-06-24T00:00:00Z",
         "budgetBreakdown": [
           {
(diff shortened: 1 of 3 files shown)


jpmckinney:
@duncandewhurst Do we update the example.json and example.md, or not?

duncandewhurst:
I've updated `example.json`, I don't think we need to reference `.requestDate` explicitly in `example.md` as it will appear in the JSON excerpt anyway.

Note: I put `example.md` rather than `example.json` in the commit message, hence the force push to correct it.

I also updated the changelog.

PR #190, opened 2019-11-20
120 add prefix list

### LINKED ISSUE #120: Add public project identifier prefixes to documentation
We need a way to expose registered OC4IDS project identifier prefixes in the documentation.

Similar mechanism as for [OCDS](http://standard.open-contracting.org/latest/en/implementation/registration/)

duncandewhurst:
Closes #120


jpmckinney (commented):
I'm not sure we need an entirely new page.

* The 'register your data' content isn't really relevant in this context (maybe it can go at the end of the implementation journey, if anywhere?).
* The step to request a prefix is already on the identifiers page, and can maybe be reworked after #191 is merged

So, it's really just about adding an iframe. Can do just add that as a section on the identifiers page?

duncandewhurst:
I've added a commit to do that. We can check if further updates are required once #191 is merged.

jpmckinney:
@duncandewhurst I added a commit. Feel free to merge if all good.

duncandewhurst:
I fixed a typo (capitalization) introduced in the previous commit - good to merge once the checks have completed.

jpmckinney:
Thanks!

PR #175, opened 2019-10-29
fixes #130 add tenderers field to contractingProcessSummary/tender

pindec:
fixes #130

PR #193, opened 2019-11-20
182 add validation properties

### LINKED ISSUE #182: Run schema-strict and add validation properties to schema
Further to @duncandewhurst's comment  [here](https://github.com/open-contracting/infrastructure/pull/178#discussion_r340889614), I note the identifier field in the budgetBreakdown array also doesn't have minLength set to 1.

duncandewhurst:
Closes #182 

I checked all the arrays that `uniqueItems` was added to and I can't see why we should permit duplicates in any of them.

The only outstanding PR or issue which involves a schema change is #175 and `uniqueItems` is already added there, so I think this can be merged.

### DIFF
--- a/schema/project-level/project-schema.json
+++ b/schema/project-level/project-schema.json
@@ -11,7 +11,8 @@
     "id": {
       "title": "Identifier or Reference",
       "description": "A unique identifier or reference number for this infrastructure project, composed of a project identifier prefix and a local project identifier. For more information, see the [project identifier guidance](https://standard.open-contracting.org/infrastructure/{{version}}/{{lang}}/guidance/identifiers/#globally-unique-project-identifiers).",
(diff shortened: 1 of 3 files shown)


schema/project-level/project-schema.json
```diff
@@ -626,8 +640,10 @@
                 "type": [
                   "number",
                   "array"
-                ]
-              }
+                ],
+                "uniqueItems": true
```
jpmckinney: We might want to allow coordinates to repeat (notably, to close a polygon).

duncandewhurst: Ah yes, in fact I think `uniqueItems` needs to be removed at both levels, since at the first level a simple point geometry could have the same latitude and longitude.

I've pushed a commit to allow both.

PR #225, opened 2020-06-10
Pull translations

### LINKED ISSUE #179: Update Spanish translation with changes in 0.9.2
Once all other issues and PRs in the 0.9.2 milestone are closed/merged, before deployment, we'll need to follow the [translation process](https://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/translation/workflow.html#steps) in the standard development handbook, to update the Spanish translation to include the changes in 0.9.2.

duncandewhurst:
Closes #179 

I had to remove the '-' from the gettext_domain_prefix in https://github.com/open-contracting/infrastructure/blob/pull-translations/docs/conf.py#L98 to get the translation to build.

I reviewed the built translation locally and noticed 3 untranslated elements:

* The sidebar header
* The copyright notice
* The CoST IDS mapping tables

It looks like this might be causing the test failure. I think something needs updating in the makefiles so that POT files are created for these resources. @jpmckinney can you advise?

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -154,10 +154,3 @@ def setup(app):
         (glob(str(basedir / 'mapping' / '*.csv')), project_build_dir, mapping_domain),
         (glob(str(basedir / 'mapping' / '*.csv')), language_dir, mapping_domain),
     ], localedir, language, mapping_headers, version=branch)
-
-    # Copy our mapping files as well. This currently does not perform translation, which would need to be added.
-    # for filename in glob(str(basedir / 'mapping' / '*.csv')):
-    #    with open(filename) as f:
-    #        mapping_file = f.read()
-    #    with open(language_dir / os.path.basename(filename), 'w') as f:
-    #        f.write(mapping_file)
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -120,7 +120,11 @@ def setup(app):
 
     language = app.config.overrides.get('language', 'en')
 
-    headers = ['Title', 'Description', 'Extension']
+    # Headers for columns to translate in codelist CSVs
+    codelist_headers = ['Title', 'Description', 'Extension']
+    # Headers for columns to translate in mapping CSVs
+    mapping_headers = ['Mapping to OC for Infrastructure','Mapping from OCDS']
+
     # The gettext domain for schema translations. Should match the domain in the `pybabel compile` command.
     schema_domain = '{}schema'.format(gettext_domain_prefix)
     # The gettext domain for codelist translations. Should match the domain in the `pybabel compile` command.
@@ -134,6 +138,7 @@ def setup(app):
 
     branch = os.getenv('TRAVIS_BRANCH', os.getenv('GITHUB_REF', 'latest').rsplit('/', 1)[-1])
 
+    # Translate schema and codelists
     translate([
         # The glob patterns in `babel_ocds_schema.cfg` should match these filenames.
         (glob(str(project_dir / '*-schema.json')), project_build_dir, schema_domain),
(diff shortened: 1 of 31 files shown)


jpmckinney:
The copyright notice isn't translated in OCDS: https://standard.open-contracting.org/latest/fr/

"Developed by" and "and" in the sidebar would be controlled by standard_theme's translations.

The mapping tables aren't translated, because they use the `csv-table-no-translate` directive. We have only used it for codelists, which get translated via `babel_ocds_codelist.cfg` and code in `conf.py`. We'd need to add something like a `babel_ocds_mapping.cfg` and similar code in `conf.py` to send the appropriate strings to Transifex for translation. The makefiles will also need updates (an extract_mappings target similar to extract_codelists, and another line in the compile target).

Alternately, we can see what Sphinx's behavior is using the `csv-table` directive. It might be more straightforward.

jpmckinney:
The test is failing because `tests/__init__.py` is configured to look for "Inicio" on the home page of the Spanish translation. That word doesn't occur, so the file needs to be configured with another word.

duncandewhurst:
Thanks @jpmckinney. I've fixed the failing test and I'm happy to leave the copyright notice and 'developed by' in the sidebar. I'm struggling to see where the translation of the project title in the sidebar header is though:

image

> Alternately, we can see what Sphinx's behavior is using the csv-table directive. It might be more straightforward.

I tested the `csv-table` directive and Sphinx extracts the individual strings from the csvs referenced in the directives.

For 0.9.2 I think we want to avoid translating the CoST IDS titles and definitions, as this will likely require some discussion, but it would be good to translate the IDS and OCDS mapping columns.

It doesn't look like Sphinx extracts anything which distinguishes which strings came from which columns in the csvs, so I can't see how we'd readily be able to do that using `csv-table`.

I'll look into creating `babel_ocds_mapping.cfg`

duncandewhurst:
I've created `babel_ocds_mapping.cfg` and made the other changes to extract just the IDS and OCDS mapping columns. The [technical implementation of translation docs](https://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/translation/implementation.html#technical-implementation-of-translation) were very helpful.

The test is failing due to the missing translations, so I've pushed the new strings to Transifex and asked Romina to translate them in CRM issue 4960

docs/conf.py
```diff
@@ -139,11 +141,14 @@ def setup(app):
...
-    for filename in glob(str(basedir / 'mapping' / '*.csv')):
-        with open(filename) as f:
-            mapping_file = f.read()
-        with open(language_dir / os.path.basename(filename), 'w') as f:
-            f.write(mapping_file)
+    # for filename in glob(str(basedir / 'mapping' / '*.csv')):
+    #    with open(filename) as f:
+    #        mapping_file = f.read()
+    #    with open(language_dir / os.path.basename(filename), 'w') as f:
+    #        f.write(mapping_file)
```
jpmckinney: Let's delete these lines, since those files are now controlled by the translation mechanism.

jpmckinney (commented):
One cleanup to do. Otherwise looks good!

If we wanted to have a hyphen for `infrastructure-schema`, etc. then we could add a hyphen at the end of `DOMAIN_PREFIX` in `config.mk` (and rename / update all references). But, not necessary.

docs/conf.py
duncandewhurst: Removed in https://github.com/open-contracting/infrastructure/pull/225/commits/caa60094f8364affe51c460d7d6cef23b54dba87

jpmckinney (approved):

PR #227, opened 2020-06-24
update version numbers to 0.9.2

duncandewhurst:


jpmckinney (approved):
Should we fix the date in the changelog?

duncandewhurst:
Ah, well spotted. I've updated the date to Monday which is when I plan to share the draft release on the mailing list.

PR #228, opened 2020-06-24
Update profile template

### LINKED ISSUE #207: Update to latest profile template
I can do this when we're ready to deploy. Updating now might interfere with the translation process.

jpmckinney:
Closes #207 

`./script/update` was pretty clean, but `./script/diff` brought up some things to update.


jpmckinney:
@duncandewhurst Once it builds, please double-check that translations of schema, codelists and mappings still work.

duncandewhurst (approved):

duncandewhurst:
When reviewing the Spanish translation I noticed some examples of codes on the codelist reference page that have been translated when they shouldn't be, e.g. 'Pre-adjudicación', 'quejas'.

This doesn't block merging this PR since the issue is present on the 0.9-dev branch too, but I'm stumped as to why this happening.

I reviewed the infrastructurecodelists MO file and the codes ('pre-award', 'complaints') don't appear individually, only as part of the code titles or descriptions. I also checked the main OCDS docs and the issue doesn't appear there.

Please could you take a look?

jpmckinney:
Aha, all the directives should be using `codelisttable`. Most codes in Transifex were copied, but some were translated. Let's see how it builds now. https://standard.open-contracting.org/infrastructure/update-profile-template/es/reference/codelists/

duncandewhurst:
Hmm now the codelists are missing altogether in both languages.

I guess because .csv is missing from the end of the filenames?

jpmckinney:
A bad search-and-replace. Let's see next.

duncandewhurst:
I think https://github.com/open-contracting/sphinxcontrib-opencontracting needs adding to the requirements

duncandewhurst:
Ah, so the codelists are displaying again but the codes are still translated.

jpmckinney:
Hmm, probably because they were extracted because csv-table was used, and their translations are in reference/codelists.po. I'll push/pull Transifex to remove them.

jpmckinney:
Hmm, I'm not maintainer on that Transifex project. Please run:

```
make
make push
make pull
```

duncandewhurst:
I'm still seeing the same issue locally. I think that may be because a '-' was added to the names of the transifex resources so they are showing as 100% untranslated. I've added you as a project maintainer.

jpmckinney:
I did `make force_push_all` to fill in those translations, then did the above. I built locally to test properly – looks like codelists page is correct now.

https://www.transifex.com/OpenDataServices/oc4ids-09/dashboard/

duncandewhurst:
Fantastic, thanks James.

PR #259, opened 2021-01-06
Fix typos and grammar issues flagged by gramma

jpmckinney:
Creating PR for @duncandewhurst's branch from 11 months ago.

### DIFF
--- a/docs/404.md
+++ b/docs/404.md
@@ -6,4 +6,4 @@
 
 We couldn't find the page you were looking for in this version of the documentation.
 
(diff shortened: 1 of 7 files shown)


jpmckinney:
@duncandewhurst How was this PR prepared? I assume we'd have to restart it against the new content.

duncandewhurst:
The commit only contains a few changes so I think restarting it rather than resolving conflicts makes sense.

I followed the instructions on https://caderek.github.io/gramma/ for checking individual files in interactive mode.

PR #300, opened 2021-03-18
Run schema-strict

### LINKED ISSUE #223: Run ocdskit schema-strict
Run ocdskit schema-strict and check that extra validations are desired.

duncandewhurst:
Closes #223 

I manually removed `uniqueItems` from `definitions/Location/properties/geometry/properties/coordinates/` and `definitions/Location/properties/geometry/properties/coordinates/items`. Otherwise, the additions look fine to me.

### DIFF
--- a/docs/reference/changelog.md
+++ b/docs/reference/changelog.md
@@ -4,7 +4,7 @@
 
 ### Documentation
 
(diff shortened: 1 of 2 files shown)


jpmckinney (approved):
I haven't reviewed, but in general this command's additions are correct.

PR #308, opened 2021-04-01
Switch to Myst Parser

### LINKED ISSUE #296: Sphinx: Specify which columns from CSV files to include or exclude
From https://github.com/open-contracting/infrastructure/pull/294#issuecomment-797765931

This would enable us to maintain one set of mapping tables but render them differently in different parts of the documentation.

I'm not sure of the best implementation approach, but options include:

* add an option to the [csv-table-no-translate](https://sphinxcontrib-opendataservices.readthedocs.io/en/latest/misc/#directive-csv-table-no-translate) directive
* make `0` an acceptable argument to the `widths` option in `csv-table-no-translate`
* reuse https://github.com/crate/sphinx_csv_filter, which has an `included_cols` option

### LINKED ISSUE #302: Switch to MyST Parser
See https://github.com/open-contracting/standard_profile_template/issues/36

Can use this script https://gist.github.com/jpmckinney/88c846651ab5411d1ed3a6fe222ce08f

Update to latest profile template as well.

jpmckinney:
closes #302, closes #296


jpmckinney:
@duncandewhurst Please review https://standard.open-contracting.org/staging/infrastructure/myst-parser/en/

You can go page-by-page in English and Spanish to check that everything renders correctly.

Some text will not be translated due to the syntax having changed and not having been updated in Spanish.

jpmckinney:
~~Hmm, not sure why it didn't build to the staging directory. Can you build and test locally?~~

jpmckinney:
It's building remotely now at above link. Fixed in https://github.com/open-contracting/deploy/commit/9b6ee8e74b145944e6c10e59316431be6c5672b7

jpmckinney:
@pindec If you are working on OC4IDS, you can also do this review (going page-by-page in English and Spanish to check that everything renders correctly).

duncandewhurst:
I checked through English and Spanish and it looks like everything renders correctly.

However, I did notice that the reactive disclosure tables were missing from the OCDS to OC4IDS mapping. It's unrelated to this PR but I've fixed it here, rather than fix it elsewhere then need to change the markdown syntax.

@jpmckinney - can you add a summary (or a link to one) of the key differences between recommonmark and MyST syntax to the handbook?

jpmckinney:
Yes, I'll update this page: https://ocds-standard-development-handbook.readthedocs.io/en/latest/meta/markdown_style_guide.html

PR #309, opened 2021-04-01
conf: Add favicon

### LINKED ISSUE #247: Favicon
I noticed that the OC4IDS standard website doesn't yet have a favicon tab icon. Is this something we should add? Firefox doesn't show a default, so it's easy to lose when a lot of tabs are open.

If we create one, I suggest that we make it a crane :)

jpmckinney:
closes #247


jpmckinney:
Favicon appears, though in Safari it seems to memorize the favicon for the domain.

PR #314, opened 2021-04-21
fixes #213, update link for releaseTag codelist

### LINKED ISSUE #213: Fix link to releaseTag codelist
The link from the [codelist reference](https://standard.open-contracting.org/infrastructure/0.9-dev/en/reference/codelists/#ocds-codelists) points to the releaseTag codelist in the OCDS documentation, however OC4IDS uses the extended releaseTag codelist from OCDS for PPPs, so the link should point to that instead.

pindec:
fixes #213


jpmckinney:
OCDS for PPPs no longer edits the release tag codelist. https://github.com/open-contracting-extensions/ocds_ppp_extension/pull/11

PR #317, opened 2021-04-23
Upgrade to OCDS 1.1.5

jpmckinney:
closes #235

After editing `borrow-schema.py`, I ran:

```
python schema/borrow-schema.py https://standard.open-contracting.org/staging/profiles/ppp/1.0-dev/en/_static/patched/
ocdskit schema-strict schema/project-level/project-schema.json
```

And then manually edited `project-schema.json` to address this issue: https://github.com/open-contracting/ocdskit/issues/172

As documented: https://ocds-standard-development-handbook.readthedocs.io/en/latest/profiles/infrastructure.html

PR #321, opened 2021-04-30
Update changelog + remove 'finalAudit' code

duncandewhurst:
@jpmckinney whilst preparing the mailing list post for OC4IDS 0.9.3 I noticed a couple of missing changelog entries and I realised that we agreed to remove the 'finalAudit' code that was added back in #317

### DIFF
--- a/schema/borrow-schema.py
+++ b/schema/borrow-schema.py
@@ -233,13 +233,11 @@ ignore = {
     'finalAudit',
     # https://github.com/open-contracting/standard/issues/870
     'contractSchedule',
-    # PPP-specific
+    # PPP-specific code or description
     'needsAssessment',
     'projectAdditionality',
     'financeAdditionality',
     'pppModeRationale',
-    'contractSchedule',
-    'riskProvisions',
     'riskComparison',
     'discountRate',
     'equityTransferCaps',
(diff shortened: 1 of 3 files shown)


schema/borrow-schema.py
```diff
@@ -273,7 +274,7 @@ def compare(actual, infra_list, ocds_list, prefix, suffix):
             # Add codes from OCDS.
             reader = csv_reader(f'{ocds_base_url}codelists/documentType.csv')
             for row in reader:
-                if row['Code'] not in seen:
+                if row['Code'] not in seen and row['Code'] != 'finalAudit':
```
jpmckinney: I tried changing the second condition to `row['Code'] not in ignore`, and that would cause 'contractSchedule' and 'riskProvisions' to also be removed from documentType.csv. Would that be the desired result? If so, I can commit it.

duncandewhurst: Hmmm, I don't recall the reason for removing 'riskProvisions', since it exists in OCDS (not just OCDS for PPPs).

Let's remove 'riskProvisions' from the `ignore` list and change the condition as you suggest.

jpmckinney: Sounds good. I checked, and 'riskProvisions' has been included since the first commit of the documentType.csv codelist: https://github.com/open-contracting/infrastructure/blame/2202f625476d24c6419c639fe69e11fa6a661221/schema/project-level/codelists/documentType.csv

jpmckinney: Ah, maybe the older logic was that OCDS for PPPs changed the definition of 'riskProvisions' in a PPP-specific way. I think the same is the case for 'needsAssessment'.

jpmckinney:
@duncandewhurst I've made the commit and updated the changelog.

duncandewhurst:
Brillant, thanks.

PR #324, opened 2021-05-05
Update template and pretranslate strings

jpmckinney:


jpmckinney:
OCDS for PPPs is being translated, so we can do another pre-translation once it's done.

duncandewhurst:
I checked a few pages and the preview looks good. Is there anything in particular that I need to check in the changed files?

Do we need to wait until the OCDS for PPPs translation is done before asking CDS to start on translating OC4IDS?

The content on the Getting Started and CoST IDS and OCDS Mapping pages is entirely new and will have little to no cross-over with OCDS for PPPs, so I'm wondering if they can make a start on those pages at least?

jpmckinney:
CDS can only really translate one project at a time, so I don't think there is any advantage to starting OC4IDS while still finishing PPPs.

jpmckinney:
Merging to avoid conflicts. We can start a new PR for the next pretranslation and translation.

PR #340, opened 2021-09-13
Update translations

duncandewhurst:
Per CRM-6221, Evelyn completed her review and all strings in Transifex are now translated and reviewed.


duncandewhurst:
@jpmckinney - any ideas why the mapping tables in https://standard.open-contracting.org/staging/infrastructure/update-translations/es/cost/ remain untranslated?

`docs/locale/es` contains `infrastructure-mappings.po` and `infrastructure-mappings.mo` files that contain the translated strings.

docs/conf.py
```diff
@@ -117,8 +117,8 @@ def setup(app):
 
     # Headers for columns to translate in codelist CSVs
     codelist_headers = ['Title', 'Description', 'Extension']
-    # Headers for columns to translate in mapping CSVs
-    mapping_headers = ['Mapping to OC for Infrastructure', 'Mapping from OCDS']
+    # Headers for columns to translate in mapping CSVs. The headers in babel_ocds_mapping.cfg  should match these
```
jpmckinney: Can also say the same about `codelist_headers` above and `babel_ocds_codelist.cfg`.

```suggestion
    # Headers for columns to translate in mapping CSVs. The headers in babel_ocds_mapping.cfg should match these.
```

jpmckinney:
> @jpmckinney - any ideas why the mapping tables in https://standard.open-contracting.org/staging/infrastructure/update-translations/es/cost/ remain untranslated?
> 
> `docs/locale/es` contains `infrastructure-mappings.po` and `infrastructure-mappings.mo` files that contain the translated strings.

Looking at just the first row of the first table, I don't see these in the PO file:

- `Project owner`
- `Name of the sponsoring Government department`
- ``Project Level: Add an `Organization` object`` ...

That's because the Babel file's `headers` need to be set before the strings are extracted. One header was changed in this PR, and the others two were never added.

duncandewhurst:
Ah, I see. We should add a prompt to https://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/translation/technical.html#extract-strings-to-translate-into-pot-files so we don't miss that step in the future.

The 'CoST IDS element' and 'CoST IDS draft definition' columns, to which `Project owner` or `Name of the sponsoring Government department` belong, have never been translated. I think that's because they are terms/definitions from the CoST IDS, which doesn't have a Spanish translation, as far as I'm aware.

I'm confused about how the po/mo files and https://www.transifex.com/open-contracting-partnership-1/oc4ids-09/translate/#es/infrastructure-mappings/322678036 came to be populated with strings when none of the mapping CSVs have the columns named in `headers`. It looks like they were updated when you extracted and pushed the strings back in May but perhaps those were old POT files.

In any case, it looks like we need to do some more translation. I tried following the steps in https://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/translation/technical.html#technical-processes-for-translation to extract and push the strings but I don't have permission to push to Transifex:

```
tx INFO: Pushing resource oc4ids-09.infrastructure-mappings
tx INFO: Pushing source file (build/locale/infrastructure-mappings.pot)
tx ERROR: Error received from server: Only maintainers are allowed to update the source file.
tx ERROR: Could not upload source file. You can use --skip to ignore this error and continue the execution.
tx ERROR: Error received from server: Only maintainers are allowed to update the source file.
```

Please can you either grant me permission, or extract the strings, push them to Transifex and then update CRM-7324 to ask CDS to translate them?

We should also add a note to https://ocds-standard-development-handbook.readthedocs.io/en/latest/profiles/infrastructure.html about the `make extract_mappings` command, which isn't otherwise documented in the handbook.

docs/conf.py
duncandewhurst: Done in https://github.com/open-contracting/infrastructure/pull/340/commits/6e629fbe3c197b476a98053b7087f6b0ac9e6d43

jpmckinney:
> Ah, I see. We should add a prompt to https://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/translation/technical.html#extract-strings-to-translate-into-pot-files so we don't miss that step in the future.

Done in referencing commit.

> The 'CoST IDS element' and 'CoST IDS draft definition' columns, to which `Project owner` or `Name of the sponsoring Government department` belong, have never been translated. I think that's because they are terms/definitions from the CoST IDS, which doesn't have a Spanish translation, as far as I'm aware.

I've added a comment to babel_ocds_mapping.cfg in 509b5a7
 
> I'm confused about how the po/mo files and https://www.transifex.com/open-contracting-partnership-1/oc4ids-09/translate/#es/infrastructure-mappings/322678036 came to be populated with strings when none of the mapping CSVs have the columns named in `headers`. It looks like they were updated when you extracted and pushed the strings back in May but perhaps those were old POT files.

Yes, possibly.

> In any case, it looks like we need to do some more translation. I tried following the steps in https://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/translation/technical.html#technical-processes-for-translation to extract and push the strings but I don't have permission to push to Transifex:
> 
> ```
> tx INFO: Pushing resource oc4ids-09.infrastructure-mappings
> tx INFO: Pushing source file (build/locale/infrastructure-mappings.pot)
> tx ERROR: Error received from server: Only maintainers are allowed to update the source file.
> tx ERROR: Could not upload source file. You can use --skip to ignore this error and continue the execution.
> tx ERROR: Error received from server: Only maintainers are allowed to update the source file.
> ```
> 
> Please can you either grant me permission, or extract the strings, push them to Transifex and then update [CRM-7324](https://crm.open-contracting.org/issues/7324) to ask CDS to translate them?

You're now Project Maintainer. I also pushed the strings (`make push`). CRM updated.

> We should also add a note to https://ocds-standard-development-handbook.readthedocs.io/en/latest/profiles/infrastructure.html about the `make extract_mappings` command, which isn't otherwise documented in the handbook.

`EXTRACT_TARGETS` is set to `extract_mappings`, such that it will be run by `make extract`, which is run by `make push`. There's not really any reason to run that task on its own.

duncandewhurst:
All sounds good. Thanks for the speedy turnaround!

> > We should also add a note to https://ocds-standard-development-handbook.readthedocs.io/en/latest/profiles/infrastructure.html about the `make extract_mappings` command, which isn't otherwise documented in the handbook.
> 
> `EXTRACT_TARGETS` is set to `extract_mappings`, such that it will be run by `make extract`, which is run by `make push`. There's not really any reason to run that task on its own.

Ah, I was looking at https://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/translation/technical.html#extract-strings-to-translate-into-pot-files, which lists the individual `extract_` tasks for the OCDS docs, and missed the later note that `make push` runs `make extract`.

duncandewhurst:
@jpmckinney this PR is now ready to merge, do you want to review again first?

jpmckinney (approved):

PR #305, opened 2021-03-31
Update blank.json with schema changes, add project package

pindec:
Fixes #304

PR #341, opened 2021-10-06
Update blank.json

duncandewhurst:
Run https://ocds-standard-development-handbook.readthedocs.io/en/latest/profiles/infrastructure.html#update-blank-json

### DIFF
--- a/docs/examples/blank.json
+++ b/docs/examples/blank.json
@@ -1,373 +1,358 @@
 {
-  "uri": "",
-  "publishedDate": "",
-  "version": "0.9",
-  "publisher": {
-    "name": "",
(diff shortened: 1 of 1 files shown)


duncandewhurst:
This removes the project package from `blank.json`, which was added by hand in https://github.com/open-contracting/infrastructure/pull/305.

json-schema-random can't cope with the `$ref` to `project-schema.json` in `project-package-schema.json` so I think it's best to omit the package from `blank.json`, otherwise we'd have to maintain it manually.

I checked where `blank.json` is referenced in the docs and I don't think the package is particularly relevant, so I'm going to merge this PR.

PR #362, opened 2022-08-03
348 publish in your own language

### LINKED ISSUE #348: Add a language field, provide guidance on how to publish in a language other than English
OC4IDS lacks a `language` field to declare the language of free-text fields and doesn't provide any guidance on language or translation.

Unlike OCDS, in-file translation isn't supported (`patternProperties` were removed in https://github.com/open-contracting/infrastructure/pull/52).

**Proposal**

1. Add a `language` field, using the same description and codelist from OCDS.

2. Add guidance on language and translation to cover:

* How to publish in your own language (i.e. set `language`, publish free-text fields in your own language, don't translate codes or field names)
* How to publish translations/multiple languages (as above, but publish separate files for each language)

duncandewhurst:
Closes #348 

To-do:

- [x] Add a pre-commit script to calculate a CSV from the schema, and display that CSV as a table in the docs using a Sphinx directive, so that readers can check which fields can be published in their own language
- [x] Add guidance on how to publish data in multiple languages (translations)


duncandewhurst:
@jpmckinney this PR is ready for review. There is one issue which I need your help with. The mapping sheet generated by the code below is missing all fields that appear in arrays, e.g. fields nested under `parties`, `documents`, `contractingProcesses` etc. https://github.com/open-contracting/infrastructure/blob/2b7421dcde0064797f02e330fb2a0fc7bc099894/manage.py#L184-L185

docs/guidance/language.md
```diff
@@ -0,0 +1,71 @@
...
+
+```json
+{
+  "id": "1",
+  "title": "ตัวอย่างโครงการ",
+  "type": "การก่อสร้าง"
+}
+```
+
+The following JSON snippet is not valid because "ชื่อ" and "พิมพ์" are not valid field names in OC4IDS:
```
jpmckinney: I'm going to rework the examples into the above text, because it's fine to add fields that are not in OC4IDS (additional fields). The problem is more specifically that these are translations of the original fields, and they are substituting for those fields.

docs/guidance/language.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Tabular data
+
+In order to ease access for non-English speakers, you can publish a spreadsheet or CSV file with field and code titles from an OC4IDS translation. Currently, OC4IDS is available in English and Spanish. If you would like to translate the schema to your own language, please [contact the OC4IDS Helpdesk](mailto:data@open-contracting.org).
+
+The following CSV except uses field and code titles from the Spanish translation of OC4IDS:
+
+| Identificador o Referencia | Título del Proyecto | Tipo de proyecto |
+| -------------------------- | ------------------- | ---------------- |
+| 1                          | proyecto de ejemplo | Construcción     |
```
jpmckinney: I'm not aware that Flatten Tool knows about codelists.

jpmckinney:
> The mapping sheet generated by the code below is missing all fields that appear in arrays

Fixed. You had to run `jsonref` on the schema first (this isn't well documented in OCDS Kit).

jpmckinney:
I made a small change to OCDS Kit, to get the mapping sheet rows directly, instead of writing to a file first. You'll need to upgrade to OCDS Kit 1.1.0.

docs/guidance/language.md
duncandewhurst: The changes look good.

docs/guidance/language.md
duncandewhurst: It doesn't. In the original text, Flatten tool was only introduced after the table and only field titles were mentioned in context of Flatten Tool. My thinking was that a publisher could substitute codes for titles as part of their publication pipeline, either before using Flatten Tool or, if they are using CSV format, after. Do you think that's worth explaining?

> In order to ease access for non-English speakers, you can publish a spreadsheet or CSV file with field and code titles from an OC4IDS translation. Currently, OC4IDS is available in English and Spanish. If you would like to translate the schema to your own language, please [contact the OC4IDS Helpdesk](mailto:data@open-contracting.org).
>
> The following CSV except uses field and code titles from the Spanish translation of OC4IDS:
> 
> | Identificador o Referencia | Título del Proyecto | Tipo de proyecto |
> | -------------------------- | ------------------- | ---------------- |
> | 1                          | proyecto de ejemplo | Construcción     |
> 
> You can use [Flatten Tool](https://flatten-tool.readthedocs.io/en/latest/) to generate a spreadsheet or CSV file with translated field titles; for example, the following command converts the example OC4IDS JSON file to xlsx format using field titles from the Spanish schema:

duncandewhurst:
> > The mapping sheet generated by the code below is missing all fields that appear in arrays
> 
> Fixed. You had to run `jsonref` on the schema first (this isn't well documented in OCDS Kit).

Thanks. I've made a PR to improve the docs: https://github.com/open-contracting/ocdskit/pull/189

> I made a small change to OCDS Kit, to get the mapping sheet rows directly, instead of writing to a file first. You'll need to upgrade to OCDS Kit 1.1.0.

Nice!

docs/guidance/language.md
jpmckinney: Hmm, yes, I thought that might be the case.

We start the page by saying that you can't translate field names or codes. In this section, we say that you can use field *titles* instead of names, and that titles are OK to translate.

If we say that codes are OK to translate, then – since that contradicts the earlier rule – we will need a little more content to explain why it's OK in spreadsheets/CSVs.

That said, I'm not sure (one way or the other) whether it's OK to translate codes. I think it's OK if there is a existing, reversible tool, e.g. like an OCDS Kit command that uses published translations of codelists.

In the absence of a tool, we're counting on publishers to reliably use the codelist translations.

Anyhow, this might be moot, because the only publishers we know who might translate codes aren't Spanish language.

I figure we can leave it for now and open an issue for consideration.

duncandewhurst: > I figure we can leave it for now and open an issue for consideration.

I've created an issue: https://github.com/open-contracting/infrastructure/issues/366

PR #364, opened 2022-08-15
include/common.mk: Ignore docs/_static directory in make autobuild

duncandewhurst:
Prior to this change, `make autobuild` gets stuck in an infinite loop because it detects a change in `docs/_static/project-level/project-level-completion.csv`.

### DIFF
--- a/include/common.mk
+++ b/include/common.mk
@@ -121,7 +121,7 @@ all: build_source compile $(TRANSLATIONS:.%=build.%) clean_current_lang
 
 .PHONY: autobuild
 autobuild: current_lang.en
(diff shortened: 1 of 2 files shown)


jpmckinney (commented):
`common.mk` is shared across OCDS profiles (including OC4IDS and OCDS itself). To configure it, can you add a variable to `include/config.mk` like `SPHINX_AUTOBUILD_EXTRA_ARGS`?

jpmckinney (approved):
Thanks! I made a corresponding change to the template here: https://github.com/open-contracting/standard_profile_template/commit/f5ff8e28f9e2b4b80ce6629f02ae008aceb8daef

Issue #244, opened 2020-11-23
Add classification scheme codelist with SDG and SDGTarget codes


duncandewhurst:
Similar to https://github.com/open-contracting/standard/issues/1098, [SIF SOURCE](https://public.sif-source.org/source/) tags infrastructure projects with the SDGs the project targets.

`additionalClassifications` can be used to model this data in OC4IDS, but we should add a codelist for `.scheme` so that publishers use a consistent scheme code for SDGs.

duncandewhurst:
Example (shared in CRM-6710):

```json
{
  "projects": [
    {
      "additionalClassifications": [
        {
          "description": "Gender equality",
          "id": "5",
          "scheme": "SDG"
        },
        {
          "description": "Decent work and economic growth",
          "id": "8",
          "scheme": "SDG"
        }
      ]
    }
  ]
}
```

duncandewhurst:
@jpmckinney are you happy with the proposal in this issue?

jpmckinney:
Yes, sounds fine.

odscrachel:
Hi @jpmckinney, I'm working on a PR for this issue but 'SDGTarget' does not match the current regex as it is not in camelCase. Using upper case letters for initialisms is consistent with the item classification scheme codelist: https://standard.open-contracting.org/latest/en/schema/codelists/#item-classification-so so I wanted to check if we should update the tests here or use camelCase?

jpmckinney:
The [codelist schema](https://github.com/open-contracting/standard-maintenance-scripts/blob/main/schema/codelist-schema.json) has for the Code column:

> The value to use in OCDS data. Codes should match the letter case of external codes (e.g. ISO 4217, JSON Schema, GeoJSON, IANA Media Types) and should be camel case, otherwise.

We've made exceptions for all-caps, for sub-codes ('abc.xyz'), for codes like 'foreignBidsFromEU', etc. We can make an exception here, but I don't mind 'sdgTarget'.

odscrachel:
Great, we'll stick to 'sdgTarget'

duncandewhurst:
Closed by https://github.com/open-contracting/infrastructure/pull/369

PR #369, opened 2022-11-07
Additional classification scheme codelist

odscrachel:
As per #244

### DIFF
--- a/schema/project-level/project-schema.json
+++ b/schema/project-level/project-schema.json
@@ -657,7 +657,7 @@
       "properties": {
         "scheme": {
           "title": "Scheme",
-          "description": "The additional classification scheme, using the open [additionalClassificationsScheme](https://standard.open-contracting.org/infrastructure/{{version}}/{{lang}}/reference/codelists/#additionalclassificationsscheme) codelist.",
+          "description": "The scheme or codelist from which the classification code is taken, using the open [additionalClassificationsScheme](https://standard.open-contracting.org/infrastructure/{{version}}/{{lang}}/reference/codelists/#additionalclassificationsscheme) codelist.",
           "type": "string",
           "codelist": "additionalClassificationsScheme.csv",
           "openCodelist": true,
--- a/schema/project-level/project-schema.json
+++ b/schema/project-level/project-schema.json
@@ -73,7 +73,7 @@
     },
     "additionalClassifications": {
       "title": "Additional classifications",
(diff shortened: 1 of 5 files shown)


schema/project-level/project-schema.json
```diff
@@ -73,7 +73,7 @@
     },
     "additionalClassifications": {
       "title": "Additional classifications",
-      "description": "One or more additional project classifications may be provided to describe the social or economic focus of the project. This classification may take place against a locally developed codelist, or a globally established codelist.",
+      "description": "One or more additional project classifications, from the [additionalClassificationsScheme codelist](https://standard.open-contracting.org/infrastructure/{{version}}/{{lang}}/reference/codelists/#additionalclassificationsscheme), may be provided to describe the social or economic focus of the project.",
```
duncandewhurst: This field description doesn't need to be updated. It's sufficient to reference the codelist in the description of `Classification.scheme`.

schema/project-level/project-schema.json
```diff
@@ -657,10 +657,10 @@
       "properties": {
         "scheme": {
           "title": "Scheme",
-          "description": "The scheme or codelist from which the classification code is taken.",
-          "type": [
-            "string"
-          ],
+          "description": "The additional classification scheme, using the open [additionalClassificationsScheme](https://standard.open-contracting.org/infrastructure/{{version}}/{{lang}}/reference/codelists/#additionalclassificationsscheme) codelist.",
```
duncandewhurst: Keep the original sentence, which better explains what a scheme is, rather than simply restating the title of the field.

```suggestion
          "description": "The scheme or codelist from which the classification code is taken, using the open [additionalClassificationsScheme](https://standard.open-contracting.org/infrastructure/{{version}}/{{lang}}/reference/codelists/#additionalclassificationsscheme) codelist.",
```

duncandewhurst (changes requested):
In keeping with OCDS:

* Name the codelist `classificationScheme`
* Add a `Category` column to the codelist and set it to `purpose` since these are schemes for describing the socioeconomic purpose of the project.

Once you've made the suggested changes, you'll need to update the changelog and run `./manage.py pre-commit`.

duncandewhurst:
Related issue, but not a blocker to this PR: https://github.com/open-contracting/infrastructure/issues/373

duncandewhurst (approved):

jpmckinney (approved):

PR #370, opened 2022-11-11
299 improve schema reference

### LINKED ISSUE #299: Move content describing the use of the schema from the worked example to the schema reference
From https://github.com/open-contracting/infrastructure/pull/298#pullrequestreview-615814396

e.g.

> The parties array is used to provide details about organizations involved in the infrastructure project and their roles. Organization references elsewhere in the data refer back to entries in this section.

At the same time, we can also rephrase to use active voice.

duncandewhurst:
* Closes #299

TO-DO:

- [x] `manage.py`
  - [x] Fix lint errors
  - [x] Handle nested properties in reference lists and in `collapse` parameter of `jsonschema` directives
  - [x] Handle cases where examples of the `Value` definition don't appear in `example.json` because they don't make sense in the context of the example: 
    - [x] `forecasts/0/observations/0/value`
    - [x] `metrics/0/observations/0/value`
    - [x] `contractingProcesses/0/summary/modifications/0/oldContractValue`
    - [x] `contractingProcesses/0/summary/modifications/0/newContractValue`
- [x] Add an introductory paragraph to explain what a sub-schema is
- [x] Review manually authored text in `docs/reference/schema.md` to reduce repetition and to identify text that should be added to field descriptions 
- [x] Review text in `docs/guidance/example.md` to identify text that should be moved to `docs/reference/schema.md` or to field descriptions
- [x] Consider removing the worked example in favour of the examples in `docs/reference/schema.md`
- [x] Update PR template (added in https://github.com/open-contracting/infrastructure/pull/374) per https://github.com/open-contracting/infrastructure/pull/374#issuecomment-1322559127


duncandewhurst:
@jpmckinney please could you check that you're happy with the direction of this PR?

I reused a script that I developed for another project to generate an improved schema reference page. You can preview it here: https://standard.open-contracting.org/staging/infrastructure/299-improve-schema-reference/en/reference/schema/. The main changes are to:

* Replace 'components' terminology with 'sub-schema' as suggested in https://github.com/open-contracting/standard/issues/1347
* Reduce reference table length by collapsing all fields that reference a sub-schema 
* Add the definition/description of each sub-schema
* Add a list of links to the fields that reference each sub-schema
* Add an example for each sub-schema

The script can be run as part of `pre-commit` to keep the schema reference documentation up to date with changes to the schema. It will preserve any manually authored content at the top of the page and under the heading for each sub-schema.

duncandewhurst:
> @jpmckinney please could you check that you're happy with the direction of this PR?

@jpmckinney just checking that you saw this request - thanks!

manage.py
```diff
@@ -163,6 +164,156 @@ def compare(actual, infra_list, ocds_list, prefix, suffix):
...
+    for defn, definition in schema["definitions"].items():
+        definition["content"] = definition.get("content", [])
+
+        # Add heading
+        definition["content"].insert(0, f"### {defn}\n")
+
+        # Add description
+        definition["content"].extend([
+            f"`{defn}` is defined as:\n\n",
+            "```{jsoninclude-quote} ../../schema/project-level/project-schema.json\n",
```
jpmckinney: sphinxcontrib-opencontracting already had https://sphinxcontrib-opencontracting.readthedocs.io/en/latest/#field-description before this was added to sphinxcontrib-opendataservices.

manage.py
```diff
@@ -163,6 +164,156 @@ def compare(actual, infra_list, ocds_list, prefix, suffix):
...
+        ])
+
+        # Add examples
+        for ref in definition["references"]:
+            if ref[0] not in schema['definitions']:
+                if ref[-1] == '0':
+                    ref.pop(-1)
+
+                definition["content"].extend([
+                  "```{eval-rst}\n",
```
jpmckinney: Don't use eval-rst. Do

````
```{jsoninclude} ../../docs/examples/example.json
:jsonpointer: ...
```
````

**Review by jpmckinney — COMMENTED** [review]
I ran the command, and each time it keeps adding more content...

Edit: Fixed locally.

**Comment by jpmckinney** [comment]
* [x] I prefer https://sphinx-inline-tabs.readthedocs.io/en/latest/ (works without JS, easier to customize) to sphinx-tabs.

File: manage.py (hunk 2 above)
Comment (jpmckinney) [reply manage.py]: Fixed locally.

**Comment by jpmckinney** [comment]
The command is failing because these don't exist

* /projects/0/forecasts/0/observations/0/value
* /projects/0/metrics/0/observations/0/value
* /projects/0/contractingProcesses/0/summary/modifications/0/oldContractValue (needs to be index 2)
* /projects/0/contractingProcesses/0/summary/modifications/0/newContractValue (needs to be index 2)

**Comment by jpmckinney** [comment]
I got `make` to run.

* I added empty `value` to forecasts and metrics for now.
* `schema.md` needs to be hand-edited to correct the indices for oldContractValue and newContractValue.


Note this diff after running pre-commit:

```diff
--- a/docs/_static/i18n.csv
+++ b/docs/_static/i18n.csv
@@ -14,7 +14,7 @@ sector,Project sector,False,
 purpose,Project purpose,True,
 additionalClassifications,Additional classifications,False,
 additionalClassifications,Classification,False,
-additionalClassifications/scheme,Scheme,True,
+additionalClassifications/scheme,Scheme,False,
 additionalClassifications/id,ID,True,
 additionalClassifications/description,Description,True,
 additionalClassifications/uri,URI,False,
```

jpmckinney:
I think we can extract the "this is referenced by" logic to a directive in sphinxcontrib-opencontracting, as suggested in https://github.com/open-contracting/standard/issues/1075#issuecomment-700994903

Otherwise, I think the rest of the logic can't be generalized, as it's mostly code to add directives in a way that's fairly specific to this project (e.g. other projects might prefer a different order or set of directives).

jpmckinney:
This might be for a follow-up PR, but I think it'd be nicer to put definitions on separate pages, as suggested in https://github.com/open-contracting/standard/issues/1075#issuecomment-694979660

duncandewhurst:
Thanks for the feedback! I take the detailed review to mean that you are happy with the direction of the PR so I will work on your comments and the checklist in the PR description.

jpmckinney:
Yup! Direction is good.

duncandewhurst:
> * [x]  I prefer https://sphinx-inline-tabs.readthedocs.io/en/latest/ (works without JS, easier to customize) to sphinx-tabs.

Should we use https://sphinx-design.readthedocs.io/en/rtd-theme/tabs.html instead?

jpmckinney:
Ah, yes, even better.

duncandewhurst:
Noting that I've enabled the `colon_fence` syntax extension as recommended in https://sphinx-design.readthedocs.io/en/rtd-theme/get_started.html. However, that does now mean there's a mix of colon fences and backtick fences.

manage.py
duncandewhurst: Doh! Which would you prefer to use?

duncandewhurst:
> I think we can extract the "this is referenced by" logic to a directive in sphinxcontrib-opencontracting, as suggested in https://github.com/open-contracting/standard/issues/1075#issuecomment-700994903

@jpmckinney do you want to do that before this PR is merged or are you happy for it to live here for now?

jpmckinney:
> Noting that I've enabled the colon_fence syntax extension as recommended

This is mainly recommended so that Markdown syntax highlighters will highlight the Markdown inside the fence (when using backticks, it highlights the whole thing as if it were a code block).

If we don't need that highlighting, it'd be simpler to not enable the extension.

> @jpmckinney do you want to do that before this PR is merged or are you happy for it to live here for now?

We can just create/update issues to do it later.

duncandewhurst:
> * [ ]  Review text in `docs/guidance/example.md` to identify text that should be moved to `docs/reference/schema.md` or to field descriptions
> * [ ]  Consider removing the worked example in favour of the examples in `docs/reference/schema.md`

@jpmckinney having reviewed the content of `docs/guidance/example.md`, it largely repeats content that is already available in either the schema reference page or in the worked example file itself. Now that the examples are integrated into the schema reference documentation, I think that we should reduce repetition by:

1. Moving the content from https://standard.open-contracting.org/staging/infrastructure/299-improve-schema-reference/en/guidance/example/#overview to the [project section](https://standard.open-contracting.org/staging/infrastructure/299-improve-schema-reference/en/reference/schema/#project) of the schema reference page
2. Removing the content from https://standard.open-contracting.org/staging/infrastructure/299-improve-schema-reference/en/guidance/example/#sections onwards
3. Updating the remaining content as follows:
   * Expand the narrative summary in the second sentence to describe more of the key features of the worked example file, e.g. the nature of the contracting processes, the variations between the budget and completion values etc. 
   * Link to the schema reference page as the place to view the examples in context of the structure of the schema

Does that sound good?

jpmckinney:
I'm having trouble visualizing what the new example.md page will look like, since if Overview is moved and Sections is removed, I'm not sure what content is left to update – is it just the content at the start of the page?

duncandewhurst:
I was thinking that it would still be useful for the worked example to have a separate page from the schema reference, which could just be the content that appears before the overview heading with the updates suggested above:

image

jpmckinney:
Aha, okay - sounds fine to me!

duncandewhurst:
@jpmckinney this is now ready for review with the exception of the build errors, which relate to the links below. Any idea how to get these to work? https://github.com/open-contracting/infrastructure/blob/22529f14db9219b97fa85a48dfa63148a88da1b9/docs/guidance/example.md?plain=1#L11-L13

I ended up restructuring the worked example page a little so that is better set up for us to add other examples in the future. Happy to hear your feedback!

docs/conf.py
```diff
@@ -42,6 +42,7 @@
     'sphinxcontrib.jsonschema',
     'sphinxcontrib.opencontracting',
     'sphinxcontrib.opendataservices',
+    'sphinx_design'
```
jpmckinney:
```suggestion
    'sphinx_design',
```

manage.py
jpmckinney: I missed this earlier comment - I prefer `field-description`.

jpmckinney (commented):
I committed the fix for the links (drop the part before `project-schema.json`).

I made a few comments, but haven't read the bulk of the changes (i.e. example.md, schema.md, and manage.py).

We're at a retreat next week, so I'll try to get back to this when I return

manage.py
duncandewhurst: Done in https://github.com/open-contracting/infrastructure/pull/370/commits/e65bbf5f2863bc0cd533f87c22adb259a8f013cb

duncandewhurst:
Thanks! I've made those updates.

duncandewhurst:
@jpmckinney just checking if you would have a chance to complete your review over the next week or two? Thanks!

jpmckinney:
In the next week or two, yes!

manage.py
```diff
@@ -163,6 +164,178 @@ def compare(actual, infra_list, ocds_list, prefix, suffix):
...
 
+def get_definition_references(schema, defn, parents=None, project_schema=None, include_nested=True):
+    """
+    Recursively generate a list of JSON pointers that reference a definition in JSON schema.
+
+    :param schema: The JSON schema
+    :defn: The name of the definition
+    :parents: A list of the parents of schema
+    :project_schema: The full project schema
+    "include_nested: Whether to include nested references
```
jpmckinney:
```suggestion
    :param schema: The JSON schema
    :param defn: The name of the definition
    :param parents: A list of the parents of schema
    :param project_schema: The full project schema
    :param include_nested: Whether to include nested references
```

jpmckinney (approved):
I'll commit my Python documentation change. I haven't reviewed the code, but I figure it works, and we'll probably change it if needed when we cover https://github.com/open-contracting/sphinxcontrib-opencontracting/issues/33

PR #426, opened 2023-11-10
Add project-level transactions and add milestones to projects and contracting processes

### LINKED ISSUE #414: Project-level disbursement records (planned and actual)
The CoST IDS/OC4IDS review proposes the following data point in the climate finance extension:

> **Disbursement records:** Disbursements dates according to financial agreement versus actual disbursements dates [value, currency, date].

Where disbursements refers to monies received by the project, rather than paid out from the project to contractors (those were already implemented in https://github.com/open-contracting/infrastructure/issues/273).

Based on the research in https://github.com/open-contracting/ocds-extensions/issues/209, project financing flows typically look like this:

```mermaid
flowchart TD
  A["Funding source
    e.g. GCF"]-- "'Parent' loan
                  agreement" -->
  B["Financing party
    e.g. UNDP"]-- "Sub-loan
                   agreement" -->
  C["Project 1"] &  D["Project 2"] & E["Project 3"]
  ````

Presumably, the disbursements that are of interest are related to the sub-loans since those are specific to the individual infrastructure projects. @EvelynDinora, please correct me if I'm wrong.

According to [Denton's Guide to Project Finance](https://www.dentons.com/~/media/6a199894417f4877adea73a76caac1a5.ashx), whilst the sub-loan agreement is between the financing party and the project company, significant disbursements are usually made directly from the financing party to suppliers, rather than via the project company:

> It will usually be the case that all disbursements of the loan for construction-related payments (and indeed other significant items) will be channelled directly through the facility agent to the relevant payees. The lenders will not usually be prepared to release drawdowns direct to the project company and take the risk that these monies are not disbursed by the project company.

For example:

```mermaid
flowchart TD
  B["Financing party
    e.g. UNDP"]-- "Sub-loan
                   agreement" -->
  C["Project"]
  C -- "Contract" --> D["Supplier
                        e.g. Skanska AB"]
  B -. Disbursements .-> D
 ```
In which case, those disbursements can be modelled in the relevant contracting process's `.transactions` array (in keeping with the mapping for the existing 'disbursement records' element in the reactive disclosure section of the CoST IDS):

```json
{
  "contractingProcesses": [
    {
      "id": "1",
      "summary": {
        "transactions": [
          {
            "id": "1",
            "payer": {
              "id": "1",
              "name": "United Nations Development Programme"
            },
            "payee": {
              "id": "2",
              "name": "Skanska AB"
            },
            "date": "2024-01-01T00:00:00Z",
            "value": {
              "amount": 10000000,
              "currency": "USD"
            }
          }
        ]
      }
    }
  ]
}
```

However, this leaves two gaps when compared to the proposed data point:

1. How to model actual disbursements which *are* made directly to the project, independently of a specific contract
2. How to the model planned disbursements (the `.transactions` array is only for actual payments)

I propose:

1. Adding a project-level `.transactions` array.
2. Reusing [OCDS 1.2's approach to modelling planned payments as milestones](https://standard.open-contracting.org/staging/1.2-dev/en/guidance/map/milestones/#delivery-and-payment-data). This would involve adding a `.milestones` array at project-level (already discussed in https://github.com/open-contracting/infrastructure/issues/253) and at contracting process level.

duncandewhurst:
**Related issues**

* Closes https://github.com/open-contracting/infrastructure/issues/414

**Description**

- Add project-level transactions.
- Add milestones to projects and contracting processes.
- Add milestoneType, milestoneStatus and milestoneCode codelists.
- Update schema for latest version of OCDS for PPPs
- Add mapping and example for disbursement records

**Merge checklist**

- [x] [Log your changes](https://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/contributing.html#logging-changes)

- [x] Update the examples:
  - [x] `docs/examples/example.json`
  - [x] `docs/examples/blank.json`
- [x] Run `./manage.py pre-commit` to update `docs/_static/i8n.csv`

If you added a new definition to the schema, run `./manage.py pre-commit`.

If you added a new codelist:

  - [x] Add an entry to `docs/reference/codelists.md`


manage.py
```diff
@@ -485,6 +486,9 @@ def copy_element(name, replacements=None, root='definitions'):
         'partyRole.csv',
         'releaseTag.csv',
         'unitClassificationScheme.csv',
+        'milestoneType.csv',
+        'milestoneStatus.csv',
+        'milestoneCode.csv'
```
jpmckinney:
```suggestion
        'milestoneCode.csv',
```

manage.py
```diff
@@ -512,6 +517,8 @@ def copy_element(name, replacements=None, root='definitions'):
         'Metric',
         'Observation',
         'Transaction',
+        'Milestone',
+        'MilestoneReference'
```
jpmckinney:
```suggestion
        'MilestoneReference',
```

jpmckinney (approved):
I didn't review every sentence of sustainability.md, but looks good. Just two commas to add to make future diffs smaller.

PR #434, opened 2023-11-17
add tender.sustainability and sustainabilityStrategies.csv

### LINKED ISSUE #400: sustainability criteria: capture if sustainability was a criteria of the award decisions
This issue relates to the following CoST IDS elements proposed in the CoST IDS/OC4IDS review:

<details>

<summary> Sustainability criteria </summary>

## Sustainability criteria

**Module:** Institutional
**Indicator:** Award criteria

### Disclosure format

> Identify the presence of sustainability and non-price attributes in the award ​​criteria. 

### OC4IDS mapping

> Contracting process level: Add a `Sustainability` object to the `.summary.tender.sustainability` array and add ‘awardCriteria’ to its `.strategies` array. 

</details>

# Proposal

Add a `sustainability` array to `contractingProcesses.summary.tender` using the OCDS [Sustainability extension](https://extensions.open-contracting.org/en/extensions/sustainability/master/) model.

Add the following fields and objects:

<div dir="ltr" style="margin-left:0pt;" align="left" id="docs-internal-guid-0fb5c5e5-7fff-4d90-31c6-115c6e1b72bd">

Path | Title | Description | Type (format) | Codelist
-- | -- | -- | -- | --
tender.sustainability | Sustainability | Information about how the contracting process incorporates sustainable public procurement. | object | -
Sustainability.strategies | Strategies | The strategies used to pursue the sustainability goal(s), using the sustainabilityStrategy codelist | array of strings | sustainabilityStrategy

</div>

# Example

```json
{
  "contractingProcesses": [
    {
      "id": "1",
      "summary": {
        "tender": {
          "sustainability": [
            {
              "strategies": [
                "awardCritera"
              ]
            }
          ]
        }
      }
    }
  ]
}

```
cc @evelyndinora

odscjen:
**Related issues**
closes #400 

**Description**

the sustainability extension has `sustainability` as an array of `Sustainability` objects. But as we're only bringing in `Sustainability.strategies` I don't think we need to make `sustainability` an array here, it can be a single instance which is what I've done

**Merge checklist**

- [ ] [Log your changes](https://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/contributing.html#logging-changes)

- [x] Update the examples:
  - [ ] `docs/examples/example.json`
  - [ ] `docs/examples/blank.json`
- [x] Run `./manage.py pre-commit` to update `docs/_static/i8n.csv`

If you added a new definition to the schema, run `./manage.py pre-commit`.

If you added a new codelist:

  - [x] Add an entry to `docs/reference/codelists.md`


duncandewhurst:
> the sustainability extension has `sustainability` as an array of `Sustainability` objects. But as we're only bringing in `Sustainability.strategies` I don't think we need to make `sustainability` an array here, it can be a single instance which is what I've done
> 

Generally, we've tried to align `ContractingProcessSummary.Tender` with the `Tender` definition in OCDS so I'm inclined to think that `.sustainability` should be an array, as in the extension. @jpmckinney what do you think?

jpmckinney:
Let's keep it as an array as in OCDS for consistency and future-proofing.

odscjen:
I've changed it to an array. I agree about the future proofing but I do think having an array of objects which themselves are made of a single array is potentially confusing.

docs/reference/changelog.md
```diff
@@ -42,6 +43,7 @@
   * [#426](https://github.com/open-contracting/infrastructure/pull/426) - milestoneType
   * [#426](https://github.com/open-contracting/infrastructure/pull/426) - milestoneStatus
   * [#426](https://github.com/open-contracting/infrastructure/pull/426) - milestoneCode
+* [#434](https://github.com/open-contracting/infrastructure/pull/434) - sustainabilityStrategy
```
duncandewhurst:
```suggestion
  * [#434](https://github.com/open-contracting/infrastructure/pull/434) - sustainabilityStrategy
```

schema/project-level/project-schema.json
```diff
@@ -477,6 +477,19 @@
...
+            "sustainability": {
+              "title": "Sustainability",
+              "description": "Information about how the contracting process incorporates sustainable public procurement.",
+              "type": [
+                "array"
+              ],
+              "items": {
+                "$ref": "#/definitions/Sustainability"
+              },
+              "wholeListMerge": true,
```
duncandewhurst: OC4IDS has no merging mechanism.

```suggestion
```

duncandewhurst (changes requested):
Happy for you to request a review from James once my comments are resolved.

schema/project-level/project-schema.json
odscjen: I remember now why I added this. It's because without it the tests fail due to there being no `id` field in `sustainability` (as they're now doing after I accepted this suggestion). There's no `id.sustainability` in the OCDS extension, but it does have the `wholeListMerge` property.

So either we add the `id` in which makes it diverge from the OCDS extension, or we keep in the `wholeListMerge` property even though it doesn't actually have any meaning in OC4IDS

duncandewhurst: Hmm, I think we might need to update the tests in that case. @jpmckinney what do you think?

odscjen:
The error is a result of no `id` field, the test needs changed as discussed https://github.com/open-contracting/infrastructure/pull/438#discussion_r1403805949

jpmckinney (approved):

Issue #467, opened 2023-12-11
Update Spanish translations


jpmckinney:
About 50% translated in 0.9.4.

duncandewhurst:
The new pages in 0.9.4 were missing from the list of resources in Transifex so I followed the [steps](https://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/translation/technical.html) in the handbook to extract, map and push strings to Transifex.

After doing so, the translation is more like ~30% complete.

duncandewhurst:
@EvelynDinora, regarding the [sustainability modules documentation](https://standard.open-contracting.org/staging/infrastructure/0.9-dev/en/cost/ids/sustainability/), please confirm which parts you want to be translated to Spanish:

![sustainability_mapping](https://github.com/open-contracting/infrastructure/assets/19265814/3158976d-d7a2-433a-8934-c7b11489574a)

duncandewhurst:
I've taken the OCDS Development Handbook instructions for translators and I'm adapting them for OC4IDS in a [Google Doc](https://docs.google.com/document/d/1y6EMJZOCaPiGnWr0QT2KipCtVh1W1MOHVIJNrigiDss/edit#heading=h.ao3xzfkkkltz).

If the mappings need to be translated, I think we'll need to add some extra instructions on which parts to leave untranslated, e.g. don't translate words in backticks or single quotes. Likewise, we should explain what to do with relative URLs, e.g. if a heading in the schema reference Markdown page is translated, should the translator update the references to that heading, or is that something that we do afterwards as part of a tidy-up?

jpmckinney:
Is there a reason there is a copy for OC4IDS, rather than making changes to the original? (e.g. moving OCDS-specific content to admonitions). From the doc's change history, it looks like it's mostly changing the examples to be about OC4IDS.

duncandewhurst:
Mostly that all the links pointed to the OCDS docs and OCDS Transifex project and I think it'll be easier for the translator to follow the instructions if the links point to the project that the translator is working on. Happy to discuss how best to integrate the docs for both projects, this was just a quick route to having something the OC4IDS translator can work with.

jpmckinney:
Cool, we can just provide both links.

duncandewhurst:
Like this? (Just checking before I add them all)

> 1. Go to the list of resources ([OCDS](), [OC4IDS]())

jpmckinney:
Yup, that works.

[...]

PR #482, opened 2024-03-07
docs: Translate CoST IDS titles and definitions

duncandewhurst:
**Related issues**

* #467

**Description**

I've pushed the newly extracted strings to Transifex:

image

Not sure if this needs a changelog entry


jpmckinney (approved):

jpmckinney:
It's a build change, so no changelog entry needed.

PR #487, opened 2024-05-17
fix: Use camelCase codes

jpmckinney:
The regex in standard-maintenance-scripts had an unescaped period (for subcodes), which allowed these to pass: https://github.com/open-contracting/standard-maintenance-scripts/commit/6730be5

I think we can at minimum correct ifc-environmental-social as I don't think anyone has used it. Hopefully we can correct pre-award as well. I didn't make a changelog entry yet, as depends on approval.

----

**Merge checklist**

- [ ] [Log your changes](https://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/contributing.html#logging-changes)

- [x] Update the examples:
  - [x] `docs/examples/example.json`


odscjen (approved):
good catch, ~~all these look fine to me, good to merge once the changelog has been updated~~ 

EDIT - realised I shouldn't have approved this without checking on the use of 'pre-award' in current publications. Most of the portals don't appear to be currently working, managed to get up-to-date data from
* Sekondi - do use `contractingProcesses.summary.status` and do use 'pre-Award' (although only 5 instances of it)
* Nuevo Leon - doesn't use `contractingProcess.summary.status` field
* Jalisco - do use `contractingProcesses.summary.status` and do use 'pre-Award' (although only 1 instance of it)

In addition Ukraine, Uganda, Malawi, Honduras and Ghana do all use the `contractingProcesses.summary.status` field but I couldn't access the data to check if they use the 'pre-Award' code.

@neelima-j you may have more knowledge of this?

odscjen (commented):
This all looks fine from a technical point of view, there are at least 2 publishers currently using the `pre-Award` code and a number of others using the field but due to broken sites I couldn't check if they were using the code

duncandewhurst:
I'm happy to change ifc-environmental, but I don't think we can remove pre-Award given that it is in use and contractingProcessStatus is a closed codelist, We could add preAward and update the description of pre-Award to indicate that it shouldn't be used, assuming we don't want to formally deprecate things whilst still on major version 0.

@jpmckinney let me know your preference.

jpmckinney:
If we want to keep it, then we might as well introduce deprecation, in this case.

duncandewhurst:
@jpmckinney I've reintroduced and deprecated 'pre-award' - the tests are failing on the code and deprecation version (0.9.5) don't conform to the regexes.

PR #501, opened 2025-01-13
Fix typo in sustainability modules mapping, fix pre-commit script

duncandewhurst:
@jpmckinney, the pre-commit script was failing with an unhashable type error as a result of https://github.com/open-contracting/infrastructure/commit/778d25947378f6b3f7f89c08b13b83cf133b7393. So, this PR reverts to using lists for inclusion conditions with unhashable types. I'm not sure if there's another way around that.


jpmckinney:
Ah, yes indeed :)

jpmckinney (approved):

